PR #12, opened 2024-04-18
fix(docs): cleaned up readme

sg-s:
## changes

- [x] fixed docs to be more accurate 
- [x]  refactored tests to use schemas to generate mock responses 
- [x] some small bugfixes
- [x] support for column programmatic keys

### DIFF
--- a/src/managed_data/schema.py
+++ b/src/managed_data/schema.py
@@ -1,0 +1,65 @@
+"""this module contains classes to generate data that 
+matches the schema of real responses"""
+
+from dataclasses import dataclass, field
+from typing import Literal
+
+RowType = Literal["row", "database"]
+
+
+@dataclass
+class FieldItem:
+    """item in fields in DescribeRow"""
+
+    columnId: str = "_col:Q2mLSmW4LNGDXAXCUUa1j"
+    cellId: str = "_cell:a6rAF44Xfyf9nwaGfo6oG"
+    validationStatus: str = "valid"
+    type: str = "text"
+    value: str = "12356789"
+
+
+@dataclass
+class ColumnItem:
+    """item in cols in DescribeRow"""
+
+    id: str = "_col:generated"
+    name: str = "Order ID"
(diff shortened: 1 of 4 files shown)


src/managed_data/schema.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import Literal
+
+RowType = Literal["row", "database"]
+
+
+@dataclass
+class FieldItem:
+    """item in fields in DescribeRow"""
+
+    columnId: str = "_col:Q2mLSmW4LNGDXAXCUUa1j"
```
jonrkarr, line 14: Shouldn't the example/test values be separated out? This will make them the default values.

src/managed_data/schema.py
jonrkarr, line 1: Love make the data model more explicit and Pythonic for users. As a user, this typically makes it much easier for me to understand how to use code compared to dictionaries which are only loosely typed.

jonrkarr (approved):

src/managed_data/schema.py
sg-s: in principle, yes. the only place this is being used is in the tests, so the values themselves don't matter -- what's being checked are the keys (and later the types too). 

i've chosen to bake the defaults in here so that the code to create fake data is very sparse -- you merely have to initialize these classes

sg-s: i could replace them with things that are more explictly placeholders, e.g.:

```
columnId: str = "_col:placeholder"
```

src/managed_data/schema.py
sg-s: down the road i plan to entirely generate this schema from the source typescript files, so that we have an exact 1:1 map

src/managed_data/schema.py
jonrkarr: I've usually just left the defaults empty if there are none. 

Programmatically generating the schema sounds great. If you think this is a v2 task, add it to the v2 epic.

PR #14, opened 2024-04-19
fix(infra): better install and better sonarcloud runs

sg-s:
## changes

- [x] better install that links executable
- [x] better docs in makefile
- [x] added an abstract class called Client that actual clients inherit from 
- [x] expanded schamas with row listings 
- [x] cleaned up testing, deleted unused functions
- [x] better coverage
- [x] reconfigured sonarcloud so that it doesn't spam us with comments, and instead is triggered by a CI action

### DIFF
--- a/src/managed_data/_api.py
+++ b/src/managed_data/_api.py
@@ -2,15 +2,13 @@
 with the data API. functions here simply wrap API endpoints."""
 
 import os
-from typing import Literal, Optional, Union
+from typing import Optional, Union
 
 import requests
 from beartype import beartype
 from deeporigin.exceptions import DeepOriginException
 from deeporigin.managed_data.client import Client, DeepOriginClient
-
-# types of rows
-RowType = Literal["row", "workspace", "database"]
+from deeporigin.managed_data.schema import RowType
 
 
 def _get_default_client(client):
--- a/src/managed_data/_api.py
+++ b/src/managed_data/_api.py
@@ -7,7 +7,7 @@ from typing import Literal, Optional, Union
 import requests
 from beartype import beartype
 from deeporigin.exceptions import DeepOriginException
-from deeporigin.managed_data.client import DeepOriginClient
+from deeporigin.managed_data.client import Client, DeepOriginClient
 
 # types of rows
 RowType = Literal["row", "workspace", "database"]
@@ -26,7 +26,7 @@ def list_rows(
     parent_id: Optional[str] = None,
     row_type: Optional[RowType] = None,
     parent_is_root: Optional[bool] = None,
-    client: Optional[DeepOriginClient] = None,
+    client: Optional[Client] = None,
 ) -> list[dict]:
     """low level API that wraps the ListRows endpoint"""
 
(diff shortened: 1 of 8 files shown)


src/managed_data/client.py
```diff
@@ -11,7 +12,18 @@
 
 
 @dataclass
-class DeepOriginClient:
+class Client(ABC):
```
jonrkarr, line 15: For thought: Why add this extra layer of abstraction? What other clients do you envision?

src/managed_data/_api.py
```diff
@@ -2,15 +2,13 @@
...
 from deeporigin.exceptions import DeepOriginException
-from deeporigin.managed_data.client import DeepOriginClient
-
-# types of rows
-RowType = Literal["row", "workspace", "database"]
+from deeporigin.managed_data.client import Client, DeepOriginClient
+from deeporigin.managed_data.schema import RowType
 
 
 def _get_default_client(client):
```
jonrkarr, line 14: For thought: why separate the client class and the methods below. Why not integrate them together so the methods and data are encapsulated together into one class? In this design, `_get_default_client` wouldn't be needed, or at least would take a different form.

jonrkarr (approved):
Looks good. I'd love to do a demo when the functionality is done, maybe at the end of the week.

I shared a few ideas for thought. None of them require response or action.

src/managed_data/schema.py
```diff
@@ -32,9 +70,9 @@ class ColumnItem:
 
 
 @dataclass
-class GenericRowDescription:
-    id: str = "_row:placeholder"
-    parentId: str = "_row:placeholder"
+class GenericRowDescription(ABC):
```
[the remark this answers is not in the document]
sg-s: long term i want to dynamically generate these all of these classes from the source typesecript classes. that way we strictly enforce the schema

src/managed_data/client.py
sg-s: the only other "client" is the mock client used in testing. previously it inherited from `DeepOriginClient` and overwrote those methods, which felt a little confusing.

src/managed_data/_api.py
sg-s: 2 reasons:

- historically i started out writing everything as functions. 
- kenny then convinced me (rightly so) that i should adopt this structure so that i could swap out clients making tests easier and more transparent. 

i like the pure functional approach that exists here, and i think it requires less boilerplate. were i to do what you suggest (if i understand you correctly), then there is one class that specified the client and all the methods in `_api`. that's exactly what i wanted to avoid -- i wanted a separation of the client from the low-level functions. were i to do your suggestion, i'd have to re-implement or overload those classes in the tests. 

finally, i agree that `_get_default_client` is clunky, but that's hard to avoid given how python initializes function default arguments. i would have liked to write this:


```python
def some_low_level_func(arg, client=DeepOriginClient):
    pass 
```

but that would require re-imports to hot-swap clients (or change config variables)

PR #18, opened 2024-04-26
feat(docs): expanded docs

sg-s:
## changes

- [x] fixed a regression where index page didn't exist
- [x] added a build make target
- [x] expanded CLI to include new functionality: list-files
- [x] added how-to section
- [x] added an automated build script for docs that checks for warnings in build 
- [x] added a github action that checks for build errors in docs
- [x] added some type hints that were missing, thanks to griffe
- [x] @jonrkarr the empty files are required so that nav doesn't break on mkdocs. they will be filled out shortly 
- [x] removed author info plugin because it doesn't work well without a token


build_docs.sh
jonrkarr, line 1: Since this script is just for CI, should we put this in `.github/workflows`?

sg-s: i'll put it in `scripts` and integrate it with the makefile

src/managed_data/_api.py
```diff
@@ -77,7 +77,7 @@ def list_files(
     assigned_row_ids: Optional[list[str]] = None,
     is_unassigned: Optional[bool] = None,
     client: Optional[Client] = None,
-):
+) -> list[dict]:
```
jonrkarr, line 80: Possibly in a future issue/PR, can we use the `schema` module to type this? This will make it easier to understand the structure of the output.

src/managed_data/_api.py
```diff
@@ -110,7 +109,7 @@ def describe_database_stats(
     database_id: str,
     *,
     client: Optional[Client] = None,
-):
+) -> dict:
```
jonrkarr, line 112: Same as above -- possibly in a future issue/PR, can we use the `schema` module to type this? This will make it easier to understand the structure of the output.

build_docs.sh
sg-s: the script is not strictly for CI. now, if you do `make build-docs`, the same script is used

src/managed_data/api.py
```diff
@@ -95,7 +95,7 @@ def get_cell_data(
     row_id: str,
     column_name: str,
     client: Optional[Client] = None,
-):
+) -> Any:
```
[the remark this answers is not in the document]
sg-s: absolutely. i like to tightly type everything. the reason for this is it's actually not clear to me what the type of this is at all -- it can be anything that is contained in a cell, which is still evolving

mkdocs.yaml
```diff
@@ -37,14 +37,14 @@ extra_css:
 
 nav:
 - Getting started:
-  - Overview: overview.md
+  - Overview: index.md
```
jonrkarr, line 40: FYI, I found the alternative file name necessary to control what's displayed in the left-navigation menu. If you figured out another way, that's better.

docs/how-to/managed-data/fetch-data.md
```diff
@@ -0,0 +1,54 @@
...
+    To save this database to a CSV, use:
+
+    ```bash
+    deeporigin data cp deeporigin://<id-of-database> <local-folder>
+    ```
+
+    The `deeporigin://` prefix identifies a resource as
+    existing on Deep Origin, and the `cp` command downloads 
+    the database as a CSV and all included files are downloaded 
+    to the `<local-folder>`
```
jonrkarr, line 32: Add a period at the end of the paragraph.

jonrkarr (approved):
Looking great.

mkdocs.yaml
sg-s: i agree that this is suboptimal behavior. however, if we don't have an index, the main page is broken and clicking on the link to the main site leads to 404. i think i know how to fix this, this will come later

src/managed_data/_api.py
sg-s: let's talk about this soon -- i'm not happy with the way things are.

docs/how-to/managed-data/fetch-data.md
sg-s: fixed

src/managed_data/_api.py
jonrkarr: Perhaps I'm missing something. We should be able to use Pydantic, dataclasses, or custom classes to model the output of each function. This might require more classes.

In any case, its not critical. We can make stricter typing a goal for a second enhanced and more performant version.

PR #19, opened 2024-04-29
feat(docs): expanded docs to show auth

sg-s:
## changes

- added info on how to authenticate to docs


docs/how-to/auth.md
```diff
@@ -0,0 +1,44 @@
...
+
+![](../images/auth-code.png)
+
+On clicking the `Confirm` button, you are shown:
+
+![](../images/auth-confirm.png)
+
+and the python client will now return an access and refresh 
+token. These tokens are also cached to disk and will automatically
+be used in subsequent interactions.
```
jonrkarr, line 40: These instructions apply to the Python client. For the CLI, should we clarify that the last step is clicking the confirmation button?

sg-s: Good catch let me fix

PR #20, opened 2024-04-29
feat(infra): avoided error messages in repeated installations

jonrkarr:
Closes https://formiclabs.atlassian.net/browse/DOS-1752

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ install:
 	@source $(CURDIR)/venv/bin/activate && \
 	    pip install -e .[test,jupyter,docs] && \
 	    deactivate
-	@-mkdir --parents ~/.deeporigin
+	@-mkdir -p ~/.deeporigin
 	if [ ! -f ~/.deeporigin/deeporigin ]; then
 		@-ln -s $(CURDIR)/venv/bin/deeporigin ~/.deeporigin/deeporigin
 	fi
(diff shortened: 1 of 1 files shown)


sg-s (commented):
doesn't work for me:

```
mkdir: illegal option -- -
usage: mkdir [-pv] [-m mode] directory_name ...
make: [Makefile:32: install] Error 64 (ignored)
if [ ! -f ~/.deeporigin/deeporigin ]; then
/bin/bash: -c: line 1: syntax error: unexpected end of file
make: *** [Makefile:33: install] Error 2
```

sg-s:
IMO having errors (ignored) in a makefile is idiomatic and the logic you're putting in here is already handled by the `-` operator

jonrkarr:
> doesn't work for me:

Fixed. I assumed bash syntax would work and didn't check.

jonrkarr:
> IMO having errors (ignored) in a makefile is idiomatic and the logic you're putting in here is already handled by the `-` operator

To me, I found output confusing because it looks like an unhandled error when the program skipped creating a new directory and adding a symlink, as expected. I would be more inclined to agree if the errors were reported as warnings. Something more like:
```
1 warning was found
-----------------------
warning >> mkdir: cannot create directory ‘/home/jonrkarr/.deeporigin’: File exists
warning >> make: [Makefile:32: install] Error 1 (ignored)
warning >> ln: failed to create symbolic link '/home/jonrkarr/.deeporigin/deeporigin': File exists
warning >> make: [Makefile:33: install] Error 1 (ignored)
```

sg-s:
yeah, that's just the way make works

sg-s (approved):

PR #25, opened 2024-05-07
fix(testing): added windows tests

sg-s:
## changes

- [x] added windows tests
- [] ~~allow overriding of environment file with config file~~

## why?

[no longer valid: not doing this]

i've changed the order of operations so that config file overrides environment variables. the reason for this is because on compute benches, environment variables are baked in and never change once jupyter lab starts. so once a bench is running, it is not possible to change env variables that are read by the client. and because env variables override files (previously), it was impossible to change any settings for the client.

### DIFF
--- a/src/config/__init__.py
+++ b/src/config/__init__.py
@@ -36,11 +36,6 @@ def get_value(
     """
     value = confuse.Configuration("deep_origin", __name__)
 
-    # read configuration from environment variables
-    # we do this first so that environment variables can be
-    # overridden by the user's configuration in the config file
-    value.set_env(sep="__")
-
     # read the default configuration
     default_config_filename = os.path.join(CONFIG_DIR, "default.yml")
     value.set_file(default_config_filename, base_for_paths=True)
@@ -50,6 +45,9 @@ def get_value(
         if os.path.isfile(user_config_filename):
             value.set_file(user_config_filename, base_for_paths=True)
             break
+
+    # read configuration from environment variables
+    value.set_env(sep="__")
 
     # validate configuration
     template = {
--- a/src/config/__init__.py
+++ b/src/config/__init__.py
@@ -37,7 +37,7 @@ def get_value(
     value = confuse.Configuration("deep_origin", __name__)
 
     # read configuration from environment variables
-    # we do this first so that environment variables can be 
+    # we do this first so that environment variables can be
     # overridden by the user's configuration in the config file
     value.set_env(sep="__")
 
@@ -50,8 +50,6 @@ def get_value(
         if os.path.isfile(user_config_filename):
             value.set_file(user_config_filename, base_for_paths=True)
             break
-
-    
 
     # validate configuration
     template = {
--- a/src/config/__init__.py
+++ b/src/config/__init__.py
@@ -36,6 +36,11 @@ def get_value(
     """
     value = confuse.Configuration("deep_origin", __name__)
 
+    # read configuration from environment variables
+    # we do this first so that environment variables can be 
+    # overridden by the user's configuration in the config file
+    value.set_env(sep="__")
+
     # read the default configuration
     default_config_filename = os.path.join(CONFIG_DIR, "default.yml")
     value.set_file(default_config_filename, base_for_paths=True)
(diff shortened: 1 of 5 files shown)


tests/test_config.py
```diff
@@ -75,6 +76,7 @@ def test_env_vars(self):
         expected_value["feature_flags"]["variables"] = True
         self.assertEqual(expected_value, value)
 
+    @unittest.skipIf(sys.platform.startswith("win"), "Test skipped on Windows")
     def test_file(self):
```
jonrkarr, line 80: This doesn't work on Windows?

If we want users to be able to use Windows, the ability to use configuration is important.

jonrkarr (approved):
I left a question about Windows.

I don't think we need to prioritize Windows. Instead, I think we should do the following in another issue/PR (I created [DOS-1913](https://formiclabs.atlassian.net/browse/DOS-1913)):
- Add a couple notes to the user documentation about what isn't support for Windows
- Raise exceptions when Windows users try to use these features which explain that Windows isn't support

Functionality that isn't supported for Windows
- Reading from configuration files?
- Installing variables/secrets

tests/test_config.py
sg-s: it does not work on windows:

https://github.com/formiclabs/deeporigin-client/actions/runs/8989133804/job/24691537723

jonrkarr: Because is small, I think we should try to fix it in another ticket

The error is below. It seems like we're not releasing a handle to a file, which likely isn't really a Windows issue.
```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp6nmcyxfs'
```

The error comes from this line. This line simply cleans up a temporary file.
https://github.com/formiclabs/deeporigin-client/blob/7a4897a2a9b7c18abdc06ce441afa90931aa0535/tests/test_config.py#L120

The other errors also stem from the same file access issue. That said, I don't think we should try to make installing variables/secrets work for Windows. So those tests should be skipped.

src/config/__init__.py
```diff
@@ -46,9 +51,6 @@ def get_value(
             value.set_file(user_config_filename, base_for_paths=True)
             break
 
-    # read configuration from environment variables
-    value.set_env(sep="__")
-
```
jonrkarr, line 51: This is ok to remove because we're not nesting configuration (i.e. none of the configuration variables have type dictionary)

sg-s: removing this breaks a bunch of tests from config

sg-s: i'm rolling this back because this seems very complex to fix and we shouldn't need this if we set the env variables correctly

jonrkarr: Actually, these needs to be restored because this line does use nesting:

https://github.com/formiclabs/deeporigin-client/blob/7a4897a2a9b7c18abdc06ce441afa90931aa0535/src/config/__init__.py#L71-L77

These tests highlight this

https://github.com/formiclabs/deeporigin-client/blob/7a4897a2a9b7c18abdc06ce441afa90931aa0535/tests/test_config.py#L64

https://github.com/formiclabs/deeporigin-client/blob/7a4897a2a9b7c18abdc06ce441afa90931aa0535/tests/test_config.py#L71

jonrkarr: Why are you trying to remove this?

This simply controls the separator between levels of configuration.

Example (`__`) in `DEEP_ORIGIN_FEATURE_FLAGS__VARIABLES`.
- `DEEP_ORIGIN`: prefix defined at https://github.com/formiclabs/deeporigin-client/blob/7a4897a2a9b7c18abdc06ce441afa90931aa0535/src/config/__init__.py#L37
- `_`: prefix separator. I believe this is defined by confuse, and we can't configure it.
- `FEATURE_FLAGS`: first level config defined at https://github.com/formiclabs/deeporigin-client/blob/7a4897a2a9b7c18abdc06ce441afa90931aa0535/src/config/__init__.py#L71
- `__`: level separator defined at https://github.com/formiclabs/deeporigin-client/blob/7a4897a2a9b7c18abdc06ce441afa90931aa0535/src/config/__init__.py#L50
- `VARIABLES`: second level config defined at https://github.com/formiclabs/deeporigin-client/blob/7a4897a2a9b7c18abdc06ce441afa90931aa0535/src/config/__init__.py#L74

sg-s: yes...and i think there is a simpler workaround for the underlying issue

PR #28, opened 2024-05-10
refactor(CLI): overhauled how tokens are handled and cached

sg-s:
## changes

- [x] fixed a bug where the CLI (and python client) would fail if there were no tokens on disk. 
- [x] completely overhauled how tokens are saved and handled. 
- [x] do_api --> auth because that module is actually about authentication and tokens, not the DO API
- [x] control over whether tokens are refreshed are not
- [x] tokens are always cached to disk, meaning we don't have to worry about caching ever. 
- [x] updated docs to reflect new syntax of CLI
- [x] changes to testing: only run tests when we really need to


src/__init__.py
```diff
@@ -3,19 +3,8 @@
 import subprocess
 from pathlib import Path
 
-from .do_api import (  # noqa: F401
```
sg-s, line 6: removing this because i want to explicitly import `auth` when we interact with the auth module

src/auth.py
```diff
@@ -5,84 +5,95 @@
...
-    "remove_cached_do_api_tokens",
-    "sign_into_do_platform",
-    "refresh_access_to_do_platform",
+    "get_tokens",
+    "cache_tokens",
+    "read_cached_tokens",
+    "remove_cached_tokens",
+    "authenticate",
+    "refresh_tokens",
 ]
```
sg-s, line 20: i found these names highly confusing and i was never sure what actions these functions performed. renamed these functions to reflect what they actually do

src/auth.py
```diff
@@ -5,84 +5,95 @@
...
+def tokens_exist() -> bool:
+    """Check if the cached API tokens exist"""
+    config = get_config()
+    return os.path.isfile(config.api_tokens_filename)
+
+
 @functools.cache
-def get_do_api_tokens(verbose: bool = False) -> tuple[str, str]:
-    """Get a token for accessing the Deep Origin API
+def get_tokens(
```
sg-s, line 31: this is the primary way we interact with the auth module. in almost every case, it should be sufficient to simply `get_tokens` and simply use that. the complexities of caching, signing in, etc., are all abstracted away

sg-s (commented):
self-review

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,20 @@
...
+  build-docs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Build docs and check for errors
+        run: bash scripts/build_docs.sh
```
sg-s, line 20: moving this to its own file so that we can trigger this to run only when the docs change. this way we don't run this job when we simply update python code

src/auth.py
jonrkarr: Sounds good.

.github/workflows/docs.yaml
jonrkarr: Nice optimization

src/auth.py
jonrkarr: I'm onboard with simplifying this.

Before I was trying to limit caching to keep alive the possibility of using this in a ComputeBench shared with multiple users, at least one of whom has root access. I think this was at least premature:
- First of all, in most cases there are better ways to facilitate collaboration than to share benches. 
- Second, sharing benches is a distant goal. 
- Third, to do bench sharing properly, I think only a special class of admin users should be able to have root access, which would enable them to see other users stored keys.

src/__init__.py
jonrkarr: No strong opinions

jonrkarr (approved):

PR #31, opened 2024-05-13
fix(docs): added info about variables

sg-s:
## changes

- [x] uses caching to speed up tests
- [x] expanded docs to refer to windows usage, what doesn't work outside of compute benches

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/setup-python@v5
         id: cached_python_setup
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.10"
           cache: 'pip'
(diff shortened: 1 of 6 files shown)


.github/workflows/docs.yaml
```diff
@@ -5,15 +5,26 @@ on:
...
           fetch-depth: 0
 
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/pyproject.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
```
sg-s, line 28: we're caching install, which should (i think) allow subsequent runs to use the cache

sg-s (commented):
self-review

jonrkarr (approved):

PR #32, opened 2024-05-17
fix(CLI): fixed a bug where `deep-origin` was still being used

sg-s:
## changes

as part of the renaming of this CLI, from deep-origin to deeporigin, we forgot to change where variables were being stored. this has now been fixed.

### DIFF
--- a/src/variables/core.py
+++ b/src/variables/core.py
@@ -376,9 +376,6 @@ def enable_variable_auto_updating(
         warnings.warn(msg, FeatureNotAvailableWarning)
         return []
 
-    # Get an access token for the Deep Origin API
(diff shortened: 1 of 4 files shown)


src/variables/core.py
```diff
@@ -376,9 +376,6 @@ def enable_variable_auto_updating(
         warnings.warn(msg, FeatureNotAvailableWarning)
         return []
 
-    # Get an access token for the Deep Origin API
-    tokens = auth.get_tokens()
-
     # load the CronTab configuration
```
sg-s, line 379: removed this because this wasn't being used anywhere

sg-s (commented):
self-review

PR #37, opened 2024-06-06
Dataframe improvements

sg-s:
## changes 

- fixes a bug where empty rows were skipped
- types many columns in pandas dataframes

### DIFF
--- a/src/managed_data/api.py
+++ b/src/managed_data/api.py
@@ -315,10 +315,13 @@ def get_dataframe(
                 use_file_names=use_file_names,
                 reference_format=reference_format,
             )
-            if column["cardinality"] == "many":
-                data[column["id"]].append(value)
+            if value is None:
+                data[column["id"]].append(None)
             else:
-                data[column["id"]].extend(value)
+                if column["cardinality"] == "many":
+                    data[column["id"]].append(value)
+                else:
+                    data[column["id"]].extend(value)
 
     if return_type == "dataframe":
         # make the dataframe
@@ -367,10 +370,10 @@ def _parse_column_value(
     field = [field for field in fields if field["columnId"] == column["id"]]
 
     if len(field) == 0:
-        return [None]
+        return None
 
     if "value" not in field[0].keys():
-        return [None]
+        return None
     value = [field[0]["value"]]
 
     # special treatment for some column types
@@ -397,7 +400,7 @@ def _parse_column_value(
 
     # if there is no item
     if len(value) == 0:
-        value = [None]
+        value = None
 
     return value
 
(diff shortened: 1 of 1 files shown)


src/managed_data/api.py
```diff
@@ -298,12 +298,14 @@ def get_dataframe(
         data[column["id"]] = []
 
     for row in rows:
-        if "fields" not in row.keys():
-            continue
-
         data[row_id].append(row["hid"])
```
sg-s, line 301: this caused empty rows to be skipped

src/managed_data/api.py
```diff
@@ -298,12 +298,14 @@ def get_dataframe(
...
         data["Validation Status"].append(row["validationStatus"])
 
+        if "fields" not in row.keys():
+            for column in columns:
+                data[column["id"]].append(None)
+            continue
+
```
sg-s, line 308: if we have an empty row, we append None to all columns and hope for the best

src/managed_data/api.py
```diff
@@ -315,10 +317,13 @@ def get_dataframe(
...
-            if column["cardinality"] == "many":
-                data[column["id"]].append(value)
+            if value is None:
+                data[column["id"]].append(None)
             else:
-                data[column["id"]].extend(value)
+                if column["cardinality"] == "many":
+                    data[column["id"]].append(value)
+                else:
+                    data[column["id"]].extend(value)
```
sg-s, line 326: values can now be None, so we need to handle that case

src/managed_data/api.py
```diff
@@ -367,10 +372,10 @@ def _parse_column_value(
     field = [field for field in fields if field["columnId"] == column["id"]]
 
     if len(field) == 0:
-        return [None]
+        return None
 
     if "value" not in field[0].keys():
-        return [None]
+        return None
```
sg-s, line 378: i don't remember why i did this, it seems simpler to return None

src/managed_data/api.py
```diff
@@ -421,17 +426,36 @@ def _type_and_cleanup_dataframe(
...
+            df[col_id] = pd.to_datetime(df[col_id])
+
+        # special treatment for string columns
+        if column["type"] in ["file", "text"]:
+            df[col_id] = df[col_id].astype("string")
+
+        if column["type"] == "boolean":
+            df[col_id] = df[col_id].astype("boolean")
+
+        # special treatment of Select columns
```
sg-s, line 445: better typing. boolean is nullable.

src/managed_data/api.py
```diff
@@ -421,17 +426,36 @@ def _type_and_cleanup_dataframe(
...
 
     # rename columns
     df = df.rename(columns=column_mapper)
 
+    # special treatment for validation status
+    categories = set(df["Validation Status"])
+    df["Validation Status"] = pd.Categorical(
+        df["Validation Status"], categories=categories
+    )
+
```
sg-s, line 458: this should work for any validation status, without needing to know what they are

sg-s (commented):
self-review

src/managed_data/api.py
kennyjwilli: Why? Can a dataframe row not just be empty?

src/managed_data/api.py
```diff
@@ -421,17 +426,36 @@ def _type_and_cleanup_dataframe(
     for column in columns:
         column_mapper[column["id"]] = column["name"]
 
-    # special treatment of Select columns
     for column in columns:
         col_id = column["id"]
 
+        print(col_id)
+        print(column["type"])
```
kennyjwilli, line 433: Remove

kennyjwilli (approved):
I'd say just remove the prints and then it's good to go 👍🏻

src/managed_data/api.py
sg-s: no. we have to put something in it

src/managed_data/api.py
sg-s: goot catch, thanks!

PR #39, opened 2024-06-11
feat(client): client can now be configured without config files

sg-s:
## problem

We want to be able to use the client in contexts where:

- we don't have write access to the filesystem
- where the config file doesn't exist
- where we want to use the client to connect to different resources than specified in the config file
- we don't want to make any authentication requests, we already have a hot token

## solution

to solve all these problems, it should be possible to do something like this:

```python
from deeporigin.managed_data.client import DeepOriginClient
client = DeepOriginClient(api_url="foo", org_id="bar")
client.authenticate(token="token")
```

### DIFF
--- a/src/managed_data/client.py
+++ b/src/managed_data/client.py
@@ -1,15 +1,13 @@
 import json
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Union
+from typing import Optional, Union
 from urllib.parse import urljoin
 
 import requests
 from beartype import beartype
 from deeporigin import auth
-from deeporigin.config import get_value
 from deeporigin.exceptions import DeepOriginException
-from deeporigin.utils import _nucleus_url
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


src/managed_data/client.py
```diff
@@ -72,12 +70,33 @@ def __str__(self):
...
+
+            api_url = _nucleus_url()
+
+        if not org_id:
+            from deeporigin.config import get_value
+
+            org_id = get_value()["organization_id"]
+
+        super().__init__()
+
```
sg-s, line 93: overloaded init to pass args to constructor

src/managed_data/client.py
```diff
@@ -72,12 +70,33 @@ def __str__(self):
...
+    def authenticate(
+        self,
+        *,
+        refresh_tokens: bool = True,
+        access_token: Optional[str] = None,
+    ):
         """authenticate to DeepOrigin API
```
sg-s, line 100: the authenticate method now accepts an access token, and if provided, does nothing

sg-s (commented):
self-review

PR #42, opened 2024-06-12
feat: supports reading tokens from env

sg-s:
## changes

the client now attempts to read access tokens from the env, if it exists, before trying to read from disk or asking the user to authenticate.

### DIFF
--- a/src/auth.py
+++ b/src/auth.py
@@ -34,11 +34,15 @@ def get_tokens(
 ) -> dict:
     """Get access token for accessing the Deep Origin API
 
-    Gets tokens to access Deep Origin API. If the tokens exist
-    on disk, those tokens are used first. On first use (within
-    a python session), tokens are refreshed if refresh_tokens
-    is `True`. On subsequent uses, tokens are not refreshed
-    even if refresh_tokens is `True`.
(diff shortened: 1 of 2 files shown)


src/auth.py
```diff
@@ -47,6 +51,11 @@ def get_tokens(
         :obj:`tuple`: API access and refresh tokens
     """
 
+    if "DEEP_ORIGIN_ACCESS_TOKEN" in os.environ:
+        return dict(
+            access=os.environ["DEEP_ORIGIN_ACCESS_TOKEN"],
+        )
+
```
sg-s, line 58: check env for access token first

src/config/__init__.py
```diff
@@ -52,7 +52,7 @@ def get_value(
     # validate configuration
     template = {
         "organization_id": confuse.String(),
-        "bench_id": confuse.String(),
+        "bench_id": confuse.Optional(confuse.String()),
```
sg-s, line 55: made this optional because we don't need it for data

sg-s (commented):
self-review

kennyjwilli (approved):

PR #43, opened 2024-06-12
feat: shortcut when running in AWS lambda; read config values from env

sg-s:
## problem

when running in lambda, reading a config file fails for some unknown reason. 

## solution

we shouldn't be reading a config file anyway, introduced a shortcut to read from envs directly

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -16,12 +16,15 @@ with open(version_filename, "r") as file:
 
     if __version__ == "0.0.0.dev0":
         # in dev mode. we use git to get a "version number"
-        # that will change with tags and commits
-        process = subprocess.run(
-            ["git", "describe", "--tags", "--dirty", "--long"],
-            capture_output=True,
-            universal_newlines=True,
-            cwd=Path(__file__).parent.parent,
-        )
-        if process.returncode == 0:
-            __version__ = process.stdout.strip()
+        # that will change with tags and commits, if possible
+        try:
+            process = subprocess.run(
+                ["git", "describe", "--tags", "--dirty", "--long"],
+                capture_output=True,
+                universal_newlines=True,
+                cwd=Path(__file__).parent.parent,
+            )
+            if process.returncode == 0:
+                __version__ = process.stdout.strip()
+        except Exception:
+            pass
--- a/src/config/__init__.py
+++ b/src/config/__init__.py
@@ -34,6 +34,19 @@ def get_value(
     Returns:
         :obj:`confuse.templates.AttrDict`: configuration for the Deep Origin CLI
     """
+
+    # if we're running on AWS lambda, read config values
+    # from env and return those.
+    if os.environ.get("AWS_LAMBDA_FUNCTION_NAME"):
+        return confuse.AttrDict(
+            dict(
+                organization_id=os.environ.get("DEEP_ORIGIN_ORGANIZATION_ID"),
+                api_endpoint=os.environ.get("DEEP_ORIGIN_API_ENDPOINT"),
+                nucleus_api_route=os.environ.get("DEEP_ORIGIN_NUCLEUS_API_ROUTE"),
+            )
+        )
+
+    # read the default configuration
     value = confuse.Configuration("deep_origin", __name__)
 
     # read the default configuration


kennyjwilli (approved):

sg-s:
@jonrkarr i am equally surprised. 

however, it looks like confuse is failing at reading a config file on a read-only system, for reasons i don't understand:

https://deeporigin.slack.com/archives/C06986CSCRJ/p1718215645735529?thread_ts=1718215522.820199&cid=C06986CSCRJ

secondly, i have verified that this workaround allows this to run on AWS

sg-s:
@jonrkarr specifically this line is failing:

```
value = confuse.Configuration("deep_origin", __name__)
```

sg-s:
the suggest fix by confuse (to set `user=False`) is something we don't want, i think, because we want confuse to read from user configs

sg-s:
@jonrkarr i have an even better minimal example. running `deeporigin config` in a read-only filesystem causes the same error:

```bash
deeporigin config

Traceback (most recent call last):
  File "/var/lang/bin/deeporigin", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/var/lang/lib/python3.12/site-packages/deeporigin/cli/__init__.py", line 117, in main
    app.run()
  File "/var/lang/lib/python3.12/site-packages/cement/core/foundation.py", line 948, in run
    return_val = self.controller._dispatch()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lang/lib/python3.12/site-packages/cement/ext/ext_argparse.py", line 810, in _dispatch
    return func()
           ^^^^^^
  File "/var/lang/lib/python3.12/site-packages/deeporigin/cli/__init__.py", line 69, in config
    data = get_value()
           ^^^^^^^^^^^
  File "/var/lang/lib/python3.12/site-packages/deeporigin/config/__init__.py", line 37, in get_value
    value = confuse.Configuration("deep_origin", __name__)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lang/lib/python3.12/site-packages/confuse/core.py", line 500, in __init__
    self.read()
  File "/var/lang/lib/python3.12/site-packages/confuse/core.py", line 535, in read
    self._add_user_source()
  File "/var/lang/lib/python3.12/site-packages/confuse/core.py", line 514, in _add_user_source
    filename = self.user_config_path()
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lang/lib/python3.12/site-packages/confuse/core.py", line 507, in user_config_path
    return os.path.join(self.config_dir(), CONFIG_FILENAME)
                        ^^^^^^^^^^^^^^^^^
  File "/var/lang/lib/python3.12/site-packages/confuse/core.py", line 572, in config_dir
    os.makedirs(appdir)
  File "<frozen os>", line 215, in makedirs
  File "<frozen os>", line 225, in makedirs
OSError: [Errno 30] Read-only file system: '/root/.config'

```

sg-s:
so confuse is trying to write something to disk when you ask it to just read the config

sg-s:
this bit of code is to blame:

https://github.com/beetbox/confuse/blob/5001e185ade4a27c613d35904147af19931690e8/confuse/core.py#L573

sg-s:
found the underlying issue. on *nix like systems, confuse assumes that ~/.config is writeable or creatable:

https://github.com/beetbox/confuse/blob/5001e185ade4a27c613d35904147af19931690e8/confuse/util.py#L9

this is not true on the lambda

jonrkarr:
Interesting. Its sort of a bug. `makedirs` isn't necessary; confuse could be implemented differently.

Two things about the new code are weird to me
- The if condition focuses on whether the environment is a lambda function, rather than the core issue whether the configuration files are writable
- The if condition returns a different data structure

One more reusable way to do this is to read from `/tmp`, which is writable in lambda. This would avoid lambda-specific code and enable `get_config` to always return the same data structure.
- If the user home directory isn't writable, read from `/tmp` instead
- If `/tmp` isn't writable, let confuse raise an error

PR #46, opened 2024-06-20
Support for uploading data 

sg-s:
## changes

- [x] ability to upload data from the client
- [x] ability to upload/set data from the CLI
- [x] tests for above
- [x] docs for above
- [x] improvements to get_dataframe -- rows now sorted

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ chosen_tests=""
 test: 
 	source $(CURDIR)/venv/bin/activate && \
 		interrogate -c pyproject.toml -v . -f 100 && \
-		python3 -m coverage run -m pytest --failed-first -k $(chosen_tests) --client $(client) && \
+		python3 -m coverage run -m pytest -x --failed-first -k $(chosen_tests) --client $(client) && \
 		python3 -m coverage html && \
 		deactivate
 	-open htmlcov/index.html
--- a/docs/how-to/managed-data/upload-data.md
+++ b/docs/how-to/managed-data/upload-data.md
@@ -9,7 +9,7 @@ This page describes how to upload data to Deep Origin Data.
     To upload a file to Deep Origin Data, use the `upload` command:
 
     ```bash
-    data upload /path/to/test.fasta 
+    deeporigin data upload /path/to/test.fasta 
     ```
 
     This will upload a file to the "Staging Area", and this
--- a/src/cli/data.py
+++ b/src/cli/data.py
@@ -397,6 +397,61 @@ databases to CSV files.
                 transpose=True,
             )
 
+    @cement.ex(
+        help="Set data in database cell",
+        arguments=[
+            (
+                ["data"],
+                {"help": "Data to set", "action": "store"},
+            ),
+            (
+                ["--database"],
+                {
+                    "type": str,
+                    "required": True,
+                    "metavar": "<database_id>",
+                    "help": "ID of database to write to",
+                },
+            ),
+            (
+                ["--column"],
+                {
+                    "type": str,
+                    "required": True,
+                    "metavar": "<column_id>",
+                    "help": "ID of column to write to",
+                },
+            ),
+            (
+                ["--row"],
+                {
+                    "type": str,
+                    "required": True,
+                    "metavar": "<row_id>",
+                    "help": "ID of row to write to",
+                },
+            ),
+            (
+                ["--json"],
+                {
+                    "action": "store_true",
+                    "help": "Whether to return JSON formatted data [default: False]",
+                },
+            ),
+        ],
+    )
+    def write(self):
+        """Write data to a cell in a database"""
+
+        api.set_cell_data(
+            self.app.pargs.data,
+            database_id=self.app.pargs.database,
+            column_id=self.app.pargs.column,
+            row_id=self.app.pargs.row,
+        )
+
+        print(f"✔︎ Wrote {self.app.pargs.data} to database")
+
 
 CONTROLLERS = [
     DataController,
--- a/src/cli/data.py
+++ b/src/cli/data.py
@@ -398,7 +398,7 @@ databases to CSV files.
             )
 
     @cement.ex(
-        help="Set data in database cell",
+        help="Write data to database cell",
         arguments=[
             (
                 ["data"],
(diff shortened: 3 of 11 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@ chosen_tests=""
 test: 
 	source $(CURDIR)/venv/bin/activate && \
 		interrogate -c pyproject.toml -v . -f 100 && \
-		python3 -m coverage run -m pytest --failed-first -k $(chosen_tests) --client $(client) && \
+		python3 -m coverage run -m pytest -x --failed-first -k $(chosen_tests) --client $(client) && \
 		python3 -m coverage html && \
```
sg-s, line 17: makes it easier to debug, this is local only

docs/how-to/managed-data/upload-data.md
```diff
@@ -9,7 +9,7 @@ This page describes how to upload data to Deep Origin Data.
     To upload a file to Deep Origin Data, use the `upload` command:
 
     ```bash
-    data upload /path/to/test.fasta 
+    deeporigin data upload /path/to/test.fasta
```
sg-s, line 12: missing deeporigin

src/cli/data.py
```diff
@@ -397,6 +397,61 @@ def upload(self):
...
+        """Write data to a cell in a database"""
+
+        api.set_cell_data(
+            self.app.pargs.data,
+            database_id=self.app.pargs.database,
+            column_id=self.app.pargs.column,
+            row_id=self.app.pargs.row,
+        )
+
+        print(f"✔︎ Wrote {self.app.pargs.data} to database")
```
sg-s, line 453: choosing not to show the actual response here because it's very complex, and the user doesn't need any of that info -- they just want to know if it worked

src/managed_data/_api.py
```diff
@@ -38,86 +38,14 @@ def _get_default_client(client: Optional[Client] = None):
...
-    if n_rows < 1:
-        raise DeepOriginException(
-            f"n_rows must be at least 1. However, n_rows was {n_rows}"
-        )
-
-    data = dict(
-        databaseId=database_id,
-        rows=[{"row": {}} for _ in range(n_rows)],
-    )
-
```
sg-s, line 120: removed these functions because they don't belong here -- they are high level

src/managed_data/api.py
```diff
@@ -13,6 +14,86 @@
...
+        raise DeepOriginException(
+            f"n_rows must be at least 1. However, n_rows was {n_rows}"
+        )
+
+    data = dict(
+        databaseId=database_id,
+        rows=[{"row": {}} for _ in range(n_rows)],
+    )
+
+    return _api.ensure_rows(data, client=client)
```
sg-s, line 46: this is now a high level function, and calls a low-level function 

this restores the pattern where high level functions don't know anything about requests, or about the client

src/managed_data/api.py
```diff
@@ -145,6 +226,101 @@ def get_cell_data(
...
+    uses the EnsureRows API endpoint"""
+
+    # first, get the type of the column
+    response = _api.describe_row(database_id, client=client)
+
+    column = [
+        col
+        for col in response["cols"]
+        if col["id"] == column_id or col["name"] == column_id or col["key"] == column_id
+    ]
```
sg-s, line 248: finding the column corresponding to the one we want

src/managed_data/api.py
```diff
@@ -145,6 +226,101 @@ def get_cell_data(
...
+        raise DeepOriginException(
+            f"Could not find column {column_id} in database {database_id}"
+        )
+
+    column = column[0]
+    if column["type"] == "select":
+        if isinstance(value, list):
+            pass
+        elif isinstance(value, str):
+            value = [value]
```
sg-s, line 260: nucleus expects lists, even for single items

src/managed_data/api.py
```diff
@@ -145,6 +226,101 @@ def get_cell_data(
...
+        elif value is None:
+            value = [""]
```
sg-s, line 262: passing an empty string unsets, but i think it's nice to be able to also pass None

src/managed_data/api.py
```diff
@@ -145,6 +226,100 @@ def get_cell_data(
...
+                raise DeepOriginException(
+                    f"Expected every item to be in the options list. However, `{item}` is not in {options}"
+                )
+
+        validated_value = dict(selectedOptions=value)
+    elif column["type"] == "text":
+        validated_value = str(value)
+    elif column["type"] == "integer":
+        try:
+            validated_value = int(value)
```
sg-s, line 280: this allows floats to be cast to int, which i think is good

src/managed_data/api.py
```diff
@@ -464,6 +639,14 @@ def _type_and_cleanup_dataframe(
...
     # doesn't actually exist in the database
     df["Validation Status"].attrs = dict(type="Validation Status")
+
+    # sort by primary key
+    primary_key = df.attrs["primary_key"]
+    df["numeric_id"] = df[primary_key].apply(lambda x: int(x.split("-")[1]))
+    df = df.sort_values(by="numeric_id")
+    df = df.drop(columns="numeric_id")
+    df.reset_index(drop=True, inplace=True)
+
```
sg-s, line 649: this complex construction is needed to make sure that db-10 is listed after db-3, for example

src/managed_data/schema.py
```diff
@@ -26,6 +26,7 @@
     "reference",
     "editor",
     "float",
+    "boolean",
```
sg-s, line 29: was missing a type

tests/mock_client.py
```diff
@@ -74,8 +74,8 @@ def invoke_list_rows(self, data: dict):
             ]
             rows += [
                 ListRowsResponse(
-                    hid=f"row-placeholder-{idx}",
-                    id=f"row-placeholder-{idx}",
+                    hid=f"row-{idx}",
+                    id=f"row-{idx}",
```
sg-s, line 78: having two `-` was breaking the logic in get_dataframe that sorted by indices. we know that indices will always have just one `-`, so this construction matches that

sg-s (commented):
self-review

jonrkarr (approved):

PR #47, opened 2024-06-21
feat(docs): updated theme to new brand identity

jonrkarr:
image
image

Closes [DOS-2680](https://deeporiginbio.atlassian.net/browse/DOS-2680)


sg-s (approved):
love it! 

later on, i wonder if we can get a more sophisitcated landing page like this:

https://squidfunk.github.io/mkdocs-material/

PR #50, opened 2024-07-02
fix: dataframe better matches DB

sg-s:
## changes

- [x] ID column is called ID in dataframe
- [x] ID column is the index in dataframe
- [x] fixed coverage to exclude test files
- [x] upgraded DeepOriginException to pretty print errors, including suggestions on how to fix it
- [x] `deeporigin config` can now show, save and load configs. 
- [x] refactored config.get_value() to only work with one file location, because having more than one was needlessly complex and was unused. 
- [x] slightly improved coverage (will improve after switching to low-level generated code)

## examples/screenshots

```
❯ deeporigin config load prod
✔︎ Configuration loaded from /Users/srinivas/.deeporigin/prod.yml
```

pretty errors:

```
❯ deeporigin data list
╭──────────────────────────────────────────────────────────────────────╮
│ Invalid configuration                                                │
├──────────────────────────────────────────────────────────────────────┤
│ The Deep Origin CLI and python client requires a valid configuration │
│ file. This field is not valid:  organization_id: must be a string    │
├──────────────────────────────────────────────────────────────────────┤
│ To fix, run `deeporigin config set organization_id <value>`          │
╰──────────────────────────────────────────────────────────────────────╯
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,13 +10,23 @@ repo=$(shell basename $(CURDIR))
 client="mock"
 chosen_tests=""
 
+
 test: 
+ifeq ($(client), "mock")
+	source $(CURDIR)/venv/bin/activate && \
+		interrogate -c pyproject.toml -v . -f 100 && \
+		python3 -m coverage run -m pytest -x --failed-first -k $(chosen_tests) --client $(client) && \
+		python3 -m coverage html && \
+		deactivate
+	-open htmlcov/index.html
+else 
 	source $(CURDIR)/venv/bin/activate && \
 		interrogate -c pyproject.toml -v . -f 100 && \
 		python3 -m coverage run -m pytest -x --failed-first -k $(chosen_tests) --client $(client) -n auto && \
 		python3 -m coverage html && \
 		deactivate
 	-open htmlcov/index.html
+endif 
 
 # set up jupyter dev kernel
 jupyter:
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ chosen_tests=""
 test: 
 	source $(CURDIR)/venv/bin/activate && \
 		interrogate -c pyproject.toml -v . -f 100 && \
-		python3 -m coverage run -m pytest -x --failed-first -k $(chosen_tests) --client $(client) && \
+		python3 -m coverage run -m pytest -x --failed-first -k $(chosen_tests) --client $(client) -n auto && \
 		python3 -m coverage html && \
 		deactivate
 	-open htmlcov/index.html
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,8 @@ universal = true
 
 [tool.coverage.run]
 omit = [
-    "src/managed_data/client.py"
+    "src/managed_data/client.py",
+    "tests/*.py",
 ]
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
     "beartype",
     "tabulate",
     "filetype",
-    "pytest-xdist",
 ]
 dynamic = ["version"]
 
@@ -36,6 +35,7 @@ test = [
     "parameterized",
     "coverage",
     "interrogate",
+    "pytest-xdist",
 ]
 jupyter = ["ipykernel","jupyter-black"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "beartype",
     "tabulate",
     "filetype",
+    "pytest-xdist",
 ]
 dynamic = ["version"]
 
--- a/src/cli/__init__.py
+++ b/src/cli/__init__.py
@@ -5,12 +5,11 @@ import warnings
 import cement
 import termcolor
 from deeporigin import auth
-from deeporigin.config import get_value
-from deeporigin.utils import _print_dict
 
 from .. import __version__
 from ..exceptions import DeepOriginException
 from ..warnings import DeepOriginWarning
+from .config import CONTROLLERS as CONFIG_CONTROLLERS
 from .context import CONTROLLERS as CONTEXT_CONTROLLERS
 from .data import CONTROLLERS as DATA_CONTROLLERS
 from .variables import CONTROLLERS as VARIABLE_CONTROLLERS
@@ -51,27 +50,6 @@ class BaseController(cement.Controller):
         """list the columns of the row and their values, where applicable"""
         auth.get_tokens(refresh=False)
 
-    @cement.ex(
-        help="Show configuration",
-        arguments=[
-            (
-                ["--json"],
-                {
-                    "action": "store_true",
-                    "help": "Whether to return JSON formatted data [default: False]",
-                },
-            ),
-        ],
-    )
-    def config(self):
-        """list the columns of the row and their values, where applicable"""
-
-        data = get_value()
-
-        data.pop("list_bench_variables_query_template", None)
-
-        _print_dict(data, json=self.app.pargs.json)
-
 
 class App(cement.App):
     class Meta:
@@ -82,6 +60,7 @@ class App(cement.App):
             + VARIABLE_CONTROLLERS
             + CONTEXT_CONTROLLERS
             + DATA_CONTROLLERS
+            + CONFIG_CONTROLLERS
         )
 
 
--- a/src/cli/config.py
+++ b/src/cli/config.py
@@ -125,7 +125,10 @@ Show and modify configuration file to connect to Deep Origin
         print(f"✔︎ Configuration loaded from {file_to_load}")
 
         # loading a config should remove api_tokens to trigger a re-authentication
-        os.remove(os.path.expanduser("~/.deeporigin/api_tokens"))
+        try:
+            os.remove(os.path.expanduser("~/.deeporigin/api_tokens"))
+        except FileNotFoundError:
+            pass
 
 
 CONTROLLERS = [
--- a/src/cli/config.py
+++ b/src/cli/config.py
@@ -65,7 +65,7 @@ Show and modify configuration file to connect to Deep Origin
         if key not in TEMPLATE.keys():
             raise DeepOriginException(
                 message=f"{key} is not a valid key for a configuration file.",
-                fix=f" Should be one of {", ".join(list(TEMPLATE.keys()))}",
+                fix=f" Should be one of {', '.join(list(TEMPLATE.keys()))}",
             )
 
         # check if config file exists
--- a/src/cli/config.py
+++ b/src/cli/config.py
@@ -88,7 +88,7 @@ Show and modify configuration file to connect to Deep Origin
         ],
     )
     def save(self):
-        """download or upload files or databases"""
+        """save a config file for later use"""
 
         name = self.app.pargs.name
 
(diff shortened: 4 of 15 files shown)


pyproject.toml
```diff
@@ -85,7 +85,8 @@ universal = true
 
 [tool.coverage.run]
 omit = [
-    "src/managed_data/client.py"
+    "src/managed_data/client.py",
+    "tests/*.py",
```
sg-s, line 89: tests shouldn't be in coverage.

src/cli/__init__.py
```diff
@@ -51,27 +50,6 @@ def authenticate(self):
...
-    )
-    def config(self):
-        """list the columns of the row and their values, where applicable"""
-
-        data = get_value()
-
-        data.pop("list_bench_variables_query_template", None)
-
-        _print_dict(data, json=self.app.pargs.json)
-
```
sg-s, line 74: refactor -- this has been moved to its own module

src/cli/config.py
```diff
@@ -0,0 +1,133 @@
...
+        _print_dict(data, json=self.app.pargs.json)
+
+    @cement.ex(
+        help="Set configuration value in config file",
+        arguments=[
+            (["key"], {"help": "Key to set", "action": "store"}),
+            (["value"], {"help": "Value to set", "action": "store"}),
+        ],
+    )
+    def set(self):
```
sg-s, line 58: this allows us to set a config value directly.

src/cli/config.py
```diff
@@ -0,0 +1,133 @@
...
+
+        print(f"✔︎ {key} → {value}")
+
+    @cement.ex(
+        help="Save configuration file for later use",
+        arguments=[
+            (["name"], {"help": "Name to save as", "action": "store"}),
+        ],
+    )
+    def save(self):
```
sg-s, line 90: this allows us to save a configration for later use. this is saved to a different file that isn't read or overwritten

src/config/__init__.py
```diff
@@ -19,17 +19,41 @@
...
+    "list_bench_variables_query_template": confuse.String(),
+    "api_tokens_filename": confuse.Filename(),
+    "variables_cache_filename": confuse.Filename(),
+    "auto_install_variables_filename": confuse.Filename(),
+    "feature_flags": confuse.Optional(
+        confuse.MappingTemplate(
+            {
+                "variables": confuse.TypeTemplate(bool),
+            }
+        )
```
sg-s, line 46: moved this out of the function -- this is now a module constant

src/config/__init__.py
```diff
@@ -54,45 +78,26 @@ def get_value(
...
+        validated_value = value.get(TEMPLATE)
     except confuse.exceptions.ConfigTypeError as exception:
         detail = str(exception).replace("\n", "\n  ")
-        raise DeepOriginException(f"The configuration is not valid:\n  {detail}")
+        key = detail.split(":")[0].strip()
+        raise DeepOriginException(
+            title="Invalid configuration",
+            message=f"The Deep Origin CLI and python client requires a valid configuration file. This field is not valid:\n {detail}",
+            fix=f"To fix, run `deeporigin config set {key} <value>`",
+        )
```
sg-s, line 101: this error is shown to the user in a pretty fashion, with suggestions on how to fix it

src/exceptions.py
```diff
@@ -1,7 +1,47 @@
...
+                tabulate(
+                    printout,
+                    tablefmt="rounded_grid",
+                ),
+                "red",
+            )
+            + "\n"
+        )
+
+        super().__init__(self.message)
```
sg-s, line 47: pretty printing of errors, so that users can see only actionable information

src/managed_data/_api.py
```diff
@@ -49,54 +48,6 @@ def ensure_rows(
...
-            params=params,
-            data=file,
-        )
-
-        if put_response.status_code != 200:
-            raise DeepOriginException("Error uploading file")
-
-    return response["file"]
-
-
```
sg-s, line 99: refactor -- all non-trivial API calls are moved to high level api

tests/mock_client.py
```diff
@@ -204,12 +204,32 @@ def invoke_list_database_rows(self, data):
...
+                    "type": "row",
+                    "dateCreated": "2024-07-03 13:54:09.888684",
+                    "dateUpdated": "2024-07-03 13:54:09.888684",
+                    "createdByUserDrn": "drn:identity::user:google-apps|srinivas@deeporigin.com",
+                    "hid": "dna-12",
+                    "validationStatus": "invalid",
+                }
+            ]
+        }
+
```
sg-s, line 223: this test will get better when we switch to generated code

tests/test_managed_data.py
```diff
@@ -67,12 +67,20 @@ def config(pytestconfig):
 
 def test_make_database_rows(config):
     with pytest.raises(DeepOriginException, match="must be at least 1"):
-        data = api.make_database_rows(
+        api.make_database_rows(
```
sg-s, line 70: data was unused

tests/test_managed_data.py
```diff
@@ -286,10 +294,6 @@ def test_get_dataframe(config):
         "Validation Status" in df.columns
     ), f"Expected to find a column called `Validation Status` in the dataframe. Instead, the columns in this dataframe are: {df.columns}"
 
-    assert (
-        df.attrs["primary_key"] in df.columns
-    ), "Expected to find the primary key as a column"
-
```
sg-s, line 292: the primary key is no longer a column, it's the index

src/managed_data/api.py
```diff
@@ -450,6 +501,11 @@ def get_dataframe(
     # figure out the rows
     rows = _api.list_database_rows(database_id, client=client)
 
+    # filter out template rows
+    rows = [
+        row for row in rows if not ("isTemplate" in row.keys() and row["isTemplate"])
+    ]
+
```
sg-s, line 508: addresses bug reported by @akash-guru

src/managed_data/api.py
```diff
@@ -458,12 +514,12 @@ def get_dataframe(
 
     columns = response["cols"]
     database_id = response["id"]
-    row_id = response["hidPrefix"] + "-id"
+    row_id = "ID"
```
sg-s, line 517: now, index is ID as in DB

src/managed_data/api.py
```diff
@@ -458,12 +514,12 @@ def get_dataframe(
...
 
     # make a dictionary with all data in the database
     data = dict()
-    data["Validation Status"] = []
     data[row_id] = []
+    data["Validation Status"] = []
```
sg-s, line 522: rearranged columns

src/managed_data/api.py
```diff
@@ -645,7 +701,7 @@ def _type_and_cleanup_dataframe(
     df["numeric_id"] = df[primary_key].apply(lambda x: int(x.split("-")[1]))
     df = df.sort_values(by="numeric_id")
     df = df.drop(columns="numeric_id")
-    df.reset_index(drop=True, inplace=True)
+    df = df.set_index("ID")
```
sg-s, line 704: the ID is now the index

sg-s (commented):
self-review

kennyjwilli (approved):
I just looked at the managed_data stuff - looks good! I'll need to update the AI codegen.

Makefile
```diff
@@ -10,13 +10,23 @@ repo=$(shell basename $(CURDIR))
...
 		deactivate
 	-open htmlcov/index.html
+else 
+	source $(CURDIR)/venv/bin/activate && \
+		interrogate -c pyproject.toml -v . -f 100 && \
+		python3 -m coverage run -m pytest -x --failed-first -k $(chosen_tests) --client $(client) -n auto && \
+		python3 -m coverage html && \
+		deactivate
+	-open htmlcov/index.html
+endif
```
sg-s, line 29: we now use xdist to run tests in parallel when running against a live instance

sg-s: thanks to this, we go from 1 min -> ~15 seconds testing against a live instance

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "beartype",
     "tabulate",
     "filetype",
+    "pytest-xdist",
```
jonrkarr, line 29: This is a test dependency.

We can separate test dependencies into a distinct section of this file
```
[project.optional-dependencies]
test = ["pytest-xdist"]
```

src/config/__init__.py
```diff
@@ -54,45 +78,26 @@ def get_value(
     value.set_file(default_config_filename, base_for_paths=True)
 
     # read configuration overrides from the user
-    for user_config_filename in user_config_filenames:
```
jonrkarr, line 57: I had configured it this way to provide the user two options (see below). This is not critical to preserve.
- Use a file in their current directory (e.g., a per-project configuration)
- Use a file in their home directory (e.g., a per-user configuration)

src/cli/config.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # check if config file exists
+        if os.path.isfile(CONFIG_YML_LOCATION):
+            with open(CONFIG_YML_LOCATION, "r") as file:
+                data = yaml.safe_load(file)
+            data[key] = value
+        else:
+            # no file.
+            data = {key: value}
+        with open(CONFIG_YML_LOCATION, "w") as file:
```
jonrkarr, line 79: Before L79, ideally we would validate the new configuration by running `src.config.get_value`. It might be helpful to refactor this function to make this easier.

src/cli/config.py
```diff
@@ -0,0 +1,136 @@
...
+            yaml.dump(data, file, default_flow_style=False)
+
+        print(f"✔︎ {key} → {value}")
+
+    @cement.ex(
+        help="Save configuration file for later use",
+        arguments=[
+            (["name"], {"help": "Name to save as", "action": "store"}),
```
jonrkarr, line 87: Rather than asking the user for a name, I think its more straightforward to ask for a file path.

src/cli/config.py
```diff
@@ -0,0 +1,136 @@
...
+    def save(self):
+        """save a config file for later use"""
+
+        name = self.app.pargs.name
+
+        # check if config file exists
+        if not os.path.isfile(CONFIG_YML_LOCATION):
+            raise DeepOriginException(
+                "Cannot save configuration for later use because no user configuration exists."
+            )
```
jonrkarr, line 99: Rather than make a copy of the user's configuration file (if it exists), would it be better to have an "export" method that exports the current configuration value?

src/cli/config.py
```diff
@@ -0,0 +1,136 @@
...
+
+        save_location = os.path.expanduser(f"~/.deeporigin/{name}.yml")
+        shutil.copy(CONFIG_YML_LOCATION, save_location)
+        print(f"✔︎ Configuration saved to {save_location}")
+
+    @cement.ex(
+        help="Save configuration file for later use",
+        arguments=[
+            (["name"], {"help": "Name to save as", "action": "store"}),
```
jonrkarr, line 108: Same as [above](https://github.com/deeporiginbio/deeporigin-client/pull/50/files#r1668006953), I think it would be simpler to ask for a file path, rather than a name.

src/cli/config.py
```diff
@@ -0,0 +1,136 @@
...
+        help = "Show and modify configuration"
+        description = """
+Show and modify configuration file to connect to Deep Origin 
+            """
+
+    def _default(self):
+        self.app.pargs.json = False
+        self.show()
+
+    @cement.ex(
```
jonrkarr, line 30: Another small function that would be helpful to add is a function that prints out the path to the user configuration file.

jonrkarr (approved):
I added several comments. These are minor, and could be addressed in subsequent PRs.

src/cli/config.py
sg-s: implemented the latter

src/cli/config.py
sg-s: it would, i'll punt this to a future fix.

src/config/__init__.py
sg-s: i have abolished `user_config_filenames` entirely in favour of a single config file located at ~/.deepirigin/config.yml

i did so because it is very confusing as to where config values are coming from when you have more than one config file, and i found myself wondering why config wasn't changing when i changed it in the file, only to realize there was a second file that was being silently used to ignore the file i was edititing. 

it also makes testing harder, and involves a fair bit of extra code.

src/cli/config.py
sg-s: fixed

src/cli/config.py
sg-s: fixed (called it profile)

pyproject.toml
sg-s: fixed

PR #54, opened 2024-07-12
refactor: switched to low-level generated code

sg-s:
## changes

- [x] consumes low-level stainless-generated code
- [x] all existing tests pass against a live instance 
- [x] dynamic code gen of simple wraps of low level code

## yet to do

- [x] all tests pass using a mocked client 
- [ ] docs rewritten to point to new code
- [ ] get_dataframe ignores columns where `systemType = ‘bodyDocument’`
- [x] refactored mocked tests to work with new client arch

## technical discussion 

please see comments inline


src/managed_data/_api.py
```diff
@@ -1,23 +1,47 @@
...
+import inspect
+import sys
+
+from deeporigin import auth
+from deeporigin.utils import _get_method
+from deeporigin_data import DeeporiginData
+
+# this dict allows us to simply wrap endpoints in the low-level
+# API so that we can more easily call them, without needing
+# to explicitly write functions for them
```
kennyjwilli, line 10: This approach feels a bit brittle and strange, but I don't have the context of _why_. Perhaps worth explaining why you took this path in a code comment would be useful.

src/managed_data/api.py
```diff
@@ -9,15 +9,125 @@
...
+# this import is to allow us to use functions
+# not marked in __all__ in _api
 from deeporigin.managed_data import _api
-from deeporigin.managed_data.client import Client
-from deeporigin.managed_data.schema import DatabaseReturnType, IDFormat
-from deeporigin.utils import PREFIXES
 
+# this import is to make sure that all simply-wrapped
+# functions in _api are available in api (this module)
+from deeporigin.managed_data._api import *  # noqa: F403
```
kennyjwilli, line 19: If I understand correctly, this re-exports functions from `_api` to this file? If so, I understand the sentiment, but, personally, I would prefer to expose these separately. I find it useful where there is a 1-1 mapping between an HTTP API and a client library. When use API extensions get merged with the 1-1 mapping, it makes it harder for me to understand.

src/managed_data/api.py
```diff
@@ -27,7 +137,7 @@ def upload_file(
 
     content_length = os.path.getsize(file_path)
 
-    response = _api.create_file_upload_url(
+    response = create_file_upload_url(  # noqa: F405
```
kennyjwilli, line 140: What is `# noqa: F405`?

src/managed_data/api.py
```diff
@@ -499,62 +615,42 @@ def get_dataframe(
...
-                if column["cardinality"] == "many":
-                    data[column["id"]].append(value)
-                else:
-                    data[column["id"]].extend(value)
+        add_row_to_data(
+            data=data,
+            row=row,
+            columns=columns,
+            use_file_names=use_file_names,
+        )
```
kennyjwilli, line 653: Nice!

src/utils.py
```diff
@@ -1,18 +1,56 @@
...
+"""Type of a column"""
+
+DATAFRAME_ATTRIBUTE_KEYS = {
+    "file_ids",
+    "id",
+    "reference_ids",
+}
+
+
+Cardinality = Literal["one", "many"]
```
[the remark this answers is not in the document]
sg-s: huh, that's what i should do! thanks!

src/managed_data/api.py
sg-s: >  I would prefer to expose these separately. 

you can use them separately, using `...import _api`.

src/managed_data/api.py
sg-s: https://www.flake8rules.com/rules/F405.html

we're using this because we're using a * import

src/utils.py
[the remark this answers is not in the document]
sg-s: hmmm, this doesnt exist in the generated code

src/managed_data/api.py
kennyjwilli: That means there'd be 2 ways to do the same thing, right?

src/utils.py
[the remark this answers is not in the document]
kennyjwilli: It's certainly in the spec. I wonder if there's a knob we can turn to get this stuff exported.

local database.csv
jonrkarr, line 1: Is this file supposed to be here?

jonrkarr (approved):
At a high level, the concept seems good. I haven't reviewed each change in detail.

local database.csv
sg-s: fixed

PR #59, opened 2024-07-31
fix: many small fixes to python client

sg-s:
## changes:

- ability to write data to many cells at once with a single function/API call
- body document columns now not included in dataframe

### DIFF
--- a/src/data_hub/api.py
+++ b/src/data_hub/api.py
@@ -424,6 +424,58 @@ def get_cell_data(
     return data[column_name]
 
 
+def set_data_in_cells(
+    *,
+    values: list,
+    row_ids: list[str],
+    column_id: str,
+    database_id: str,
+    client=None,
+):
+    """Set data in multiple cells to some value."""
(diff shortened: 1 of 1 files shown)


src/data_hub/api.py
```diff
@@ -424,6 +424,58 @@ def get_cell_data(
     return data[column_name]
 
 
+def set_data_in_cells(
```
sg-s, line 427: this new function allows you to write data to many cells within the same column

src/data_hub/api.py
```diff
@@ -715,6 +779,14 @@ def get_dataframe(
...
     reference_ids = []
 
+    # remove body document columns because they are not
+    # shown in the UI as columns
+    columns = [
+        col
+        for col in columns
+        if "systemType" not in col.keys() or col["systemType"] != "bodyDocument"
+    ]
+
```
sg-s, line 789: ignoring body document column

PR #60, opened 2024-08-01
minor improvements to client

sg-s:
## changes

-  removed a file that shouldn't be there
- methods now have a `debug` parameter which prints debug information about every API request


src/data_hub/_api.py
```diff
@@ -111,6 +124,13 @@ def dynamic_function(*, client=None, **kwargs):
             else:
                 raise error
 
+        if isinstance(response, _response.APIResponse):
```
jonrkarr, line 127: Would `if debug:` be simpler?

The meaning of this if condition seems hard to understand. I understand the condition is when response has the type APIResponse. What type APIResponse means vs other possible types could be clearer. It seems like the condition for this is similar to `debug=True` and that this is related to `client = client.with_raw_response`.

src/data_hub/_api.py
```diff
@@ -76,6 +81,9 @@ def _get_default_client(client=None, refresh: bool = True):
             base_url=base_url,
         )
 
+        if debug:
+            client = client.with_raw_response
```
jonrkarr, line 85: This feels like a complicated way to make the behavior conditional on `debug=True`.

jonrkarr (approved):
Approving because this seems to only affect the behavior for debugging. 

I think conditioning on `if debug:` would make the code easier to understand.

src/data_hub/_api.py
sg-s: This makes sense. I'll change.

sg-s: fixed

src/data_hub/_api.py
sg-s: fixed

PR #63, opened 2024-08-05
fix: minor improvements

sg-s:
## changes

- [ ] `do data show row-id` now includes the row HID
- [ ] body document not shown by default 
- [ ] new function to read out body document

### DIFF
--- a/src/cli/data.py
+++ b/src/cli/data.py
@@ -306,6 +306,13 @@ class DataController(cement.Controller):
                     "help": "Whether to return JSON formatted data [default: False]",
                 },
             ),
+            (
+                ["--body-document"],
+                {
+                    "action": "store_true",
+                    "help": "Whether to show body document [default: False]",
+                },
+            ),
         ],
     )
     def show(self):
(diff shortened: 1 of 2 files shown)


src/cli/data.py
```diff
@@ -312,15 +312,33 @@ def describe(self):
                     "help": "Whether to return data in JSON format [default: False]",
                 },
             ),
+            (
+                ["--body-document"],
+                {
+                    "action": "store_true",
+                    "help": "Whether to show body document [default: False]",
```
jonrkarr, line 319: change "body document" to "notebook"

src/data_hub/api.py
```diff
@@ -1036,6 +1036,31 @@ def get_columns(
         return response["fields"]
 
 
+@beartype
+def get_body_document(row_id: str, *, client=None):
```
jonrkarr, line 1040: if possible, change to `get_notebook`

jonrkarr (approved):
Looks good. 

1 request, please change "body document" to "notebook" in the help strings to align our terminology.

src/data_hub/api.py
sg-s: fixed

src/cli/data.py
sg-s: fixed

PR #65, opened 2024-08-05
Updates to documentation and bugfixes

sg-s:
## changes

- [x] updated docs, clarified various pages, cleaned up some how-to docs
- [x] fixed a bug where a column called "sf" was shown
- [x] fixed a bug where long strings/values broke the table shown in the CLI

### DIFF
--- a/src/data_hub/api.py
+++ b/src/data_hub/api.py
@@ -1040,8 +1040,6 @@ def get_columns(
 def get_notebook(row_id: str, *, client=None) -> list:
     """Get the notebook of a row, if it exists
 
-
-
     Args:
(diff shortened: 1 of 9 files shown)


src/data_hub/api.py
```diff
@@ -1037,10 +1037,9 @@ def get_columns(
 
 
 @beartype
-def get_notebook(row_id: str, *, client=None):
+def get_notebook(row_id: str, *, client=None) -> list:
```
sg-s, line 1040: we need this type annotation for docs to build successfully

jonrkarr (approved):
I made a few minor changes to the docs

PR #67, opened 2024-08-07
feat: clarified and standardized error messages

jonrkarr:


src/cli/config.py
```diff
@@ -99,13 +99,14 @@ def save(self):
         name = self.app.pargs.profile
 
         # check if config file exists
-        if not os.path.isfile(CONFIG_YML_LOCATION):
-            raise DeepOriginException(
-                "Cannot save configuration for later use because no user configuration exists."
-            )
```
jonrkarr, line 105: I thought this condition isn't necessary. A non-overrrided configuration is valid.

sg-s (approved):

PR #70, opened 2024-08-13
feat: updated to new version of deleting database columns

sg-s:
## changes

@kennyjwilli made a breaking change to the API and the syntax for deleting columns now requires us to specify database ID

- [x] upgraded to new version of low-level SDK built off of the new openAPI spec
- [x] pinned that version 
- [x] updated readme
- [x] updated the CLI code

### DIFF
--- a/docs/how-to/data-hub/delete.md
+++ b/docs/how-to/data-hub/delete.md
@@ -34,7 +34,7 @@ To delete multiple rows, databases and folders, run:
 
 ## Delete database columns
 
-To delete columns in databases, specify the name of the column and the database it exists in:
+To delete columns in databases, run the following command, specifying the IDs of the columns to delete and their parent database:
 
 === "CLI"
 
--- a/docs/how-to/data-hub/delete.md
+++ b/docs/how-to/data-hub/delete.md
@@ -34,24 +34,23 @@ To delete multiple rows, databases and folders, run:
 
 ## Delete database columns
 
-To delete columns in databases, run:
+To delete columns in databases, specify the name of the column and the database it exists in:
 
 === "CLI"
 
     ```bash
-    deeporigin data delete --ids <ids> --columns
+    deeporigin data delete \
+        --ids <ids> \
+        --database <database-id> \
+        --columns
     ```
-
-    !!! warning "Column IDs"
-        Currently, columns can only be deleted using their IDs, which are distinct from their names. To view the IDs of the columns of a database, run: 
-
-        ```bash
-        deeporigin data describe <database-id>
-        ```
 
 === "Python"
 
     ```py
     from deeporigin.data_hub import api
-    api.delete_database_column(column_id="col-id")
+    api.delete_database_column(
+        column_id="col-id",
+        database_id="database-id",
+    )
     ```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
     "tabulate",
     "filetype",
     "httpx",
-    "deeporigin-data-sdk>=0.1.0a6",
+    "deeporigin-data-sdk==0.1.0a7",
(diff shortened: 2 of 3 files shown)


docs/how-to/data-hub/delete.md
```diff
@@ -34,24 +34,23 @@ To delete multiple rows, databases and folders, run:
...
+        --columns
     ```
 
-    !!! warning "Column IDs"
-        Currently, columns can only be deleted using their IDs, which are distinct from their names. To view the IDs of the columns of a database, run: 
-
-        ```bash
-        deeporigin data describe <database-id>
-        ```
-
```
sg-s, line 51: no longer true

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "tabulate",
     "filetype",
     "httpx",
-    "deeporigin-data-sdk>=0.1.0a6",
+    "deeporigin-data-sdk==0.1.0a7",
```
sg-s, line 30: we pin to this version for now

src/cli/data.py
```diff
@@ -566,6 +566,14 @@ def new(self):
                     "help": "IDs of the resources to delete",
                 },
             ),
+            (
+                ["--database"],
+                {
+                    "type": str,
+                    "required": False,
+                    "help": "IDs of database that columns are in",
```
jonrkarr, line 574: Should this be plural. as in:

> IDs of the databases that the columns are in

docs/how-to/data-hub/delete.md
```diff
@@ -34,24 +34,23 @@ To delete multiple rows, databases and folders, run:
 
 ## Delete database columns
 
-To delete columns in databases, run:
+To delete columns in databases, specify the name of the column and the database it exists in:
```
jonrkarr, line 37: I like to make instructions a little more explicit:

> To delete columns in databases, run the following command, specifying the name of the column and its parent database:

src/cli/data.py
sg-s: it is. we can delete many columns in 1 DB, but they all have to be in one DB

docs/how-to/data-hub/delete.md
sg-s: fixed

src/cli/data.py
```diff
@@ -586,9 +594,14 @@ def delete(self):
         """Delete rows, columns, databases and/or folders"""
 
         if self.app.pargs.columns:
+            if self.app.pargs.database is None:
+                raise DeepOriginException(
+                    "You must specify a database to delete columns in using --database"
```
jonrkarr, line 599: I like to provide instructions first and explanation second. How about this?

> Use the --database argument to specify the parent database. To delete a column, the parent database must be specified.

jonrkarr (changes requested):
See suggestions above

src/cli/data.py
sg-s: fixed

src/cli/data.py
sg-s: also, it *has* to be plural to delete other objects

jonrkarr: Since its one database (singular), I think "ID" singular would be clearer.

sg-s: oof, you're right. my bad. got confused with IDs of objects

jonrkarr: > also, it has to be plural to delete other objects

In that case, should it be this:

> IDs of the databases that the columns are in

sg-s: fixed

sg-s: no, you're right. it's ID singular for DB, and IDs plural for objects

jonrkarr (approved):

PR #71, opened 2024-08-15
fix: fixed some small bugs

sg-s:
## changes

see inline comments

### DIFF
--- a/src/data_hub/api.py
+++ b/src/data_hub/api.py
@@ -322,17 +322,8 @@ def assign_files_to_cell(
 
     """
 
-    if row_id is None:
-        data = make_database_rows(
-            database_id=database_id,
-            n_rows=1,
-            client=client,
-        )
-        row_id = data.rows[0].id
-
     rows = [
         {
-            "rowId": row_id,
             "row": {},
             "cells": [
                 {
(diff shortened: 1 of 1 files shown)


src/data_hub/api.py
```diff
@@ -343,6 +334,9 @@ def assign_files_to_cell(
         },
     ]
 
+    if row_id:
+        rows[0]["rowId"] = row_id
+
```
sg-s, line 339: @kennyjwilli suggested this which is simpler

src/data_hub/api.py
```diff
@@ -505,7 +499,7 @@ def set_data_in_cells(
     column = [
         col
         for col in response.cols
-        if col["id"] == column_id or col["name"] == column_id or col["key"] == column_id
```
sg-s, line 508: no more column keys

src/data_hub/api.py
```diff
@@ -568,7 +562,7 @@ def set_cell_data(
...
```
sg-s, line 571: no more column keys

kennyjwilli (approved):
Nice simplification 👍🏻

PR #72, opened 2024-08-15
feat: better way of interacting with databases

sg-s:
## problem description

a typical user flow is 

- get some data from a database. data is then exstant as a pandas dataframe
- do some computation on it
- maybe write results to a new column in the dataframe
- write those results back to the DO database (sync back)

right now, to do so, we have to worry about API calls where we have to laboriously pass database ID, column IDs, etc. 

### user experience

```python
df = api.get_dataframe("foo")

# do some compute on it
df.sync()
```

### technical implementation 

we create a new class that subclasses a pandas dataframe, and write special methods to talk to the data hub

## changes

- [ ] api.get_dataframe() stores database info in dataframe attrs
- [ ] dataframe contains information about columns 
- [ ] dataframe.sync works

### DIFF
--- a/src/data_hub/api.py
+++ b/src/data_hub/api.py
@@ -505,7 +505,7 @@ def set_data_in_cells(
     column = [
         col
         for col in response.cols
-        if col["id"] == column_id or col["name"] == column_id or col["key"] == column_id
+        if col["id"] == column_id or col["name"] == column_id
     ]
 
     if len(column) != 1:
@@ -568,7 +568,7 @@ def set_cell_data(
     column = [
         col
         for col in response.cols
-        if col["id"] == column_id or col["name"] == column_id or col["key"] == column_id
+        if col["id"] == column_id or col["name"] == column_id
     ]
 
     if len(column) != 1:
@@ -867,9 +867,9 @@ def get_dataframe(
 
         # this import is here because we don't want to
         # import pandas unless we actually use this function
-        import pandas as pd
+        from deeporigin.data_hub.dataframe import DataFrame
 
-        df = pd.DataFrame(data)
+        df = DataFrame(data)
         df.attrs["file_ids"] = list(set(file_ids))
         df.attrs["reference_ids"] = list(set(reference_ids))
         df.attrs["id"] = database_id
@@ -1170,10 +1170,7 @@ def get_row_data(
     column_name_mapper = dict()
     column_cardinality_mapper = dict()
     for col in parent_response.cols:
-        if use_column_keys:
-            column_name_mapper[col["id"]] = col["key"]
-        else:
-            column_name_mapper[col["id"]] = col["name"]
+        column_name_mapper[col["id"]] = col["name"]
         column_cardinality_mapper[col["id"]] = col["cardinality"]
 
     # now use this to construct the required dictionary
@@ -1182,10 +1179,7 @@ def get_row_data(
     for col in parent_response.cols:
         if "systemType" in col.keys() and col["systemType"] == "bodyDocument":
             continue
-        if use_column_keys:
-            row_data[col["key"]] = None
-        else:
-            row_data[col["name"]] = None
+        row_data[col["name"]] = None
     if not hasattr(response, "fields"):
         return row_data
     for field in response.fields:
--- a/src/data_hub/dataframe.py
+++ b/src/data_hub/dataframe.py
@@ -1,0 +1,36 @@
+from typing import Optional
+
+import pandas as pd
+from deeporigin.data_hub import api
+
+
+class DataFrame(pd.DataFrame):
+    def _repr_html_(self):
+        header = "<h3>Deep Origin DataFrame</h3>"
+        df_html = super()._repr_html_()
+        return header + df_html
+
+    def __repr__(self):
+        header = "Deep Origin DataFrame\n"
+        df_representation = super().__repr__()
+        return header + df_representation
+
+    def sync(self, *, columns: Optional[list] = None):
+        print("Syncing...")
+
+        for column in self.columns:
+            if column == "Validation Status":
+                continue
+
+            print(column)
+
+            api.set_data_in_cells(
+                values=self[column],
+                row_ids=list(self.index),
+                column_id=column,
+                database_id=self.attrs["id"],
+            )
+
+    @property
+    def _constructor(self):
+        return DataFrame


docs/ref/data-hub/types.md
```diff
@@ -1,6 +1,15 @@
-# Types and constants
```
sg-s, line 1: this doc page is no longer about types and constants, so rephrased

mkdocs.yaml
```diff
@@ -44,6 +44,8 @@ nav:
   - Authenticate: how-to/auth.md
 - Data hub: 
   - data-hub/index.md
+  - Deep Origin DataFrames:
+    - Tutorial: data-hub/dataframes.md
   - How to:
```
sg-s, line 49: i intend this to be the primary way users interact with databases, so made it front and center

src/data_hub/api.py
```diff
@@ -861,14 +861,22 @@ def get_dataframe(
 
         # this import is here because we don't want to
         # import pandas unless we actually use this function
-        import pandas as pd
+        from deeporigin.data_hub.dataframe import DataFrame
```
sg-s, line 864: we no longer directly import pandas

src/data_hub/api.py
```diff
@@ -861,14 +861,22 @@ def get_dataframe(
...
 
-        return _type_and_cleanup_dataframe(df, columns)
+        df = _type_and_cleanup_dataframe(df, columns)
+
+        # if this code is running the lambda, we do not
+        # turn on auto_sync by default because we don't want
+        # side effects from code that the LLM writes
+        if os.environ.get("AWS_LAMBDA_FUNCTION_NAME") is None:
+            df.auto_sync = True
+        return df
```
sg-s, line 879: really don't want to allow the LLM to accidentally modify DBs by doing something it thinks is local

src/data_hub/api.py
```diff
@@ -994,7 +1002,7 @@ def _row_to_dict(row, *, use_file_names: bool = True):
 
 @beartype
 def _type_and_cleanup_dataframe(
-    df,  # pd.Dataframe, not typed to avoid pandas import
+    df,  # Dataframe, not typed to avoid pandas import
```
sg-s, line 1005: this is a DO DataFrame

src/data_hub/api.py
```diff
@@ -1164,10 +1172,7 @@ def get_row_data(
     column_name_mapper = dict()
     column_cardinality_mapper = dict()
     for col in parent_response.cols:
-        if use_column_keys:
-            column_name_mapper[col["id"]] = col["key"]
-        else:
-            column_name_mapper[col["id"]] = col["name"]
+        column_name_mapper[col["id"]] = col["name"]
```
sg-s, line 1175: removed support for column keys

src/data_hub/api.py
```diff
@@ -1176,10 +1181,7 @@ def get_row_data(
     for col in parent_response.cols:
         if "systemType" in col.keys() and col["systemType"] == "bodyDocument":
             continue
-        if use_column_keys:
-            row_data[col["key"]] = None
-        else:
-            row_data[col["name"]] = None
+        row_data[col["name"]] = None
     if not hasattr(response, "fields"):
```
sg-s, line 1185: no column keys

src/utils.py
```diff
@@ -17,10 +19,10 @@
 
 
 RowType = Literal["row", "database", "workspace"]
-"""Type of a row"""
+"""Type of a row. In Deep Origin, a row can be a database row, a database or a workspace"""
 
 FileStatus = Literal["ready", "archived"]
-"""Status of a file"""
+"""Status of a file. Ready files are ready to be used, downloaded, and operated on."""
```
sg-s, line 25: better docs

src/data_hub/api.py
kennyjwilli: Conversely, it could be quite convenient for a user to write this code in a code executor block and expect it to work. But I do agree that this is super dangerous in general. Perhaps let's start with this safe implementation and strategize how to deal with the possibility of writes.

src/utils.py
```diff
@@ -63,6 +66,32 @@ class PREFIXES:
...
+        row_type (RowType): type of the resource
+
+    Returns:
+        str: URL for the resource
+    """
+
+    env = get_value()["env"]
+    org = get_value()["organization_id"]
+    if env == "prod":
+        url = "https://os.deeporigin.io/org/{org}/data/{row_type}/{name}"
```
kennyjwilli, line 88: Is this missing `f`?

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,159 @@
...
+            if rows is None:
+                # we're updating a whole column
+                rows = list(self.index)
+
+            api.set_data_in_cells(
+                values=self[column],
+                row_ids=rows,
+                column_id=column,
+                database_id=self.attrs["id"],
+            )
```
kennyjwilli, line 153: I don't know what happens behind the scenes in this call, and maybe you've already done this, but edits to 1 database can be optimized into a single HTTP request.

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,159 @@
...
+                continue
+
+            column_metadata = [
+                col for col in self.attrs["metadata"]["cols"] if col["name"] == column
+            ]
+
+            if len(column_metadata) == 0:
+                raise NotImplementedError(
+                    "Column metadata not found. This is likely because it's a new column"
+                )
```
kennyjwilli, line 137: In _theory_ could we implement automatic column creation?

kennyjwilli (approved):
SO FREAKING COOL!

src/data_hub/api.py
sg-s: they can get the "automatic" functionality to work using the `auto_sync` attribute

src/data_hub/dataframe.py
sg-s: this is a single request to EnsureRows

src/data_hub/dataframe.py
sg-s: yeah. a little tricky because we need to guess types. but possible. i'll make a ticket for this

akash-guru (commented):
Looks great!

src/data_hub/dataframe.py
sg-s: https://deeporigin.atlassian.net/browse/DOS-3660

jonrkarr (changes requested):
I live the simpler UX!

# Auto-saving
Instead of auto-saving changes back to Deep Origin, I think we should give users a method to easily save changes back to Deep Origin. To fit with the typical DataFrame UX, we could add a `to_deeporigin` method.
- I think auto-save would make it too easy for users to unintentionally overwrite their data.
- I think auto-save would be surprising to users.
- By making auto-saving the default, it makes it a little harder for users to get a data frame that they can transiently manipulate. For this use case, users would have to go an extra step, rather than making users go an extra step to sync their changes back to deeporigin.

# Generalizing to enable additional use cases
By making `to_deeporigin` more explicit, we could also build on this idea further to enable additional use cases. See below.
- Use `to_deeporigin` to create a new database. 
  1. User creates a DataFrame
  2. User calls `to_deeporigin` to sync the DataFrame to Deep Origin. The user must supply arguments for the database ID, name, and row ID prefix. `to_deeporigin` then attaches this metadata to the DataFrame.
- Use `to_deeporigin` to create a new rows

# Documentation
I think we should expand the documentation for this a little to outline how various scenarios are handled (or not). I don't think we need to handle all of the use cases right now. Rather, we could simply make it clear what is presently supported and what isn't presently supported so users aren't surprised.

- Creating new databases
- Creating new columns
- Creating new rows
- Editing the names of databases
- Editing the columns
- Deleting columns
- Deleting rows

src/data_hub/api.py
jonrkarr: > I do agree that this is super dangerous in general

Highly agree. I love the spirit of the PR. With a couple minor changes (see suggestions [below](https://github.com/deeporiginbio/deeporigin-client/pull/72#pullrequestreview-2244151789)), I think we can provide a UX which is (a) safe, (b) simpler, and (c) highly consistent with the way that DataFrames are frequently used.

docs/ref/data-hub/types.md
```diff
@@ -1,6 +1,15 @@
...
+# Classes and Constants
```
jonrkarr, line 1: Lowercase "Constants". The sections headings are generally sentence case.

mkdocs.yaml
```diff
@@ -54,12 +56,13 @@ nav:
   - API reference:
     - High-level API: ref/data-hub/high-level-api.md
     - Low-level API: ref/data-hub/low-level-api.md
-    - Types and constants: ref/data-hub/types.md
+    - Classes & Constants: ref/data-hub/types.md
```
jonrkarr, line 59: Lowercase "Constants". The sections headings are generally sentence case.

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,79 @@
...
+
+This page describes how to use Deep Origin DataFrames, which are the primary object you will use to interact with databases on Deep Origin. This page will cover:
+
+- fetching data from a Deep Origin database
+- modifying data locally
+- writing data back to Deep Origin
+
+
+!!! question "What is a Deep Origin DataFrame?"
+    A Deep Origin DataFrame is a subclass of a [pandas.DataFrame](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html) that is backed by a Deep Origin database. Because it is a subclass of a pandas DataFrame, all pandas DataFrame methods work on Deep Origin DataFrames.
```
jonrkarr, line 11: It would be helpful to elaborate a little more on what "backed" means.
- DataFrames can easily be created from Deep Origin databases
- DataFrames can easily be saved to Deep Origin databases
- If we stick with auto-saving, this would be very important to describe here. Actually, we'd want to style this as a danger callout. Auto save is too dangerous to keep.

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,79 @@
+# Using Deep Origin DataFrames
+
+This page describes how to use Deep Origin DataFrames, which are the primary object you will use to interact with databases on Deep Origin. This page will cover:
+
+- fetching data from a Deep Origin database
+- modifying data locally
+- writing data back to Deep Origin
+
+
+!!! question "What is a Deep Origin DataFrame?"
```
jonrkarr, line 10: I think we should inline this into the main narrative because this is essential to understanding how this works.

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,79 @@
...
+    A Deep Origin DataFrame is a subclass of a [pandas.DataFrame](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html) that is backed by a Deep Origin database. Because it is a subclass of a pandas DataFrame, all pandas DataFrame methods work on Deep Origin DataFrames. 
+
+## Create a DataFrame
+
+Create a DataFrame using:
+
+```python
+from deeporigin.data_hub import api
+df = api.get_data_frame("database-id")
```
jonrkarr, line 19: We could make this more pandas-like by adding a `from_deeporigin` method to DoDataFrame

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,79 @@
...
+df
+```
+
+which should show you something like this:
+
+![DataFrame](../images/dataframe.png)
+
+
+!!! tip "Information in the DataFrame"
+    In addition to information you would find in the rows and columns of a pandas DataFrame, a Deep Origin DataFrame also contains metadata about the underling database. In the view above we also see:
```
jonrkarr, line 34: "underling" is misspelled

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,79 @@
...
+## Create a DataFrame
+
+Create a DataFrame using:
+
+```python
+from deeporigin.data_hub import api
+df = api.get_data_frame("database-id")
+```
+
+In an interactive web-based environment such as Jupyter Lab, you should be able to view the DataFrame using:
```
jonrkarr, line 22: Replace 

> In an interactive web-based environment such as Jupyter Lab

with

> In a Jupyter notebook,

The pretty display depends on rendering Jupyter notebooks, rather than hosting in the web. For example, users would also experience this in VS Code.

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,79 @@
...
+
+```python
+df.at["sgs-1", "y"] = 10
+```
+
+
+## Write data back to Deep Origin
+
+!!! success "Nothing to do here!"
+    A Deep Origin DataFrame automatically syncs back to the Deep origin database. Making the changes above also makes changes in the source database.
```
jonrkarr, line 66: Instead of auto-saving, I strongly think we should add a `to_deeporigin` method to DoDataFrame.

With this change, we can then 
- Combine this section and the following on manual syncing
- Add a section on how to create a new database

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+### Manually syncing
+
+Automatic syncing occurs when the `auto_sync` attribute of a Deep Origin DataFrame is set to `True`. You may want to turn `auto_sync` off if you want to manually sync back to the source database, for example if you want to keep the changes in the source database.
+
+To manually sync changes, use the `sync` method:
+
+```python
+df.sync()
```
jonrkarr, line 76: To be more consistent with DataFrame method names, I think we should rename this to `to_deeporigin`.

jonrkarr: I have read the implementation yet. If the method were called `sync`, I'd expect a behavior like  this

1. Check that the central version of the database hasn't changed since loaded into the DoDataFrame
2. If the central version and DataFrame have both modified the same cell, raise an exception
3. Have additional options to ranges exceptions when either rows have diverged or the number of rows has diverged (new rows have been added to the central database)
4. Save the changes to the DataFrame to the central database

jonrkarr: I think the section should elaborate further on how saving is handled
- New rows
- New columns
- Renamed columns
- Deleted rows
- Deleted columns

src/data_hub/api.py
```diff
@@ -861,14 +861,22 @@ def get_dataframe(
...
         # this import is here because we don't want to
         # import pandas unless we actually use this function
-        import pandas as pd
+        from deeporigin.data_hub.dataframe import DataFrame
 
-        df = pd.DataFrame(data)
+        df = DataFrame(data)
         df.attrs["file_ids"] = list(set(file_ids))
         df.attrs["reference_ids"] = list(set(reference_ids))
         df.attrs["id"] = database_id
```
jonrkarr, line 869: Is this redundant with the information in `db_row` and therefore df.attrs["metadata"]?

src/utils.py
```diff
@@ -17,10 +19,10 @@
 
 
 RowType = Literal["row", "database", "workspace"]
```
jonrkarr, line 21: Can we choose a more general term than "row" such as "object". I think its confusing to overload "row".

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,159 @@
...
+            # object, so we need to index into the pandas object
+            if self.obj.auto_sync:
+                self.obj.sync(columns=columns, rows=rows)
+
+    @property
+    def at(self):
+        """Override the `at` property to return an AtIndexer"""
+        return self.AtIndexer(self)
+
+    def __setitem__(self, key, value):
```
jonrkarr, line 60: What about intercepting item deletion?

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,159 @@
...
+        return header + txt + df_html
+
+    def __repr__(self):
+        """method override to customize printing in an interactive session"""
+
+        header = f'{self.attrs["metadata"]["hid"]}\n'
+        df_representation = super().__repr__()
+        return header + df_representation
+
+    def sync(
```
jonrkarr, line 105: For consistency with pandas, I think we could rename this to `to_deeporigin`

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,159 @@
...
+from datetime import datetime
+from typing import Optional
+
+import humanize
+import pandas as pd
+from deeporigin.data_hub import api
+from deeporigin.utils import construct_resource_url
+
+
+class DataFrame(pd.DataFrame):
```
jonrkarr, line 16: For consistency with pandas, I think we could add a `from_deeporigin` method rather than encouraging users to use `api.get_data_frame`. 

`from_deeporigin` could be implementing using `api.get_data_frame`. It could largely be syntactic sugar for `api.get_data_frame`.

sg-s: I like the mirroring of pandas functions. Pandas has pd.read_csv, maybe we could have deeporigin.read_database?

src/data_hub/dataframe.py
sg-s: As in, deleting a whole column? Good point. Haven't thought about it.

src/data_hub/dataframe.py
jonrkarr: The DataFrame class includes `from_dict` and `from_records`.  It seems more straightforward to add methods to the DataFrame class. In that case, I think we should follow the pattern `from_*`.

`read_*` are on the pandas module.

src/data_hub/dataframe.py
jonrkarr: I don't think its a priority to support deleting columns now. Instead, we could more simply document what is and isn't supported, and leave the rest to later.

src/utils.py
sg-s: good catch!

src/data_hub/dataframe.py
sg-s: i agree.

pandas is weird here -- we're supposed to use `pd.read_csv` and not `pd.DataFrame.from_csv` [link](https://pandas.pydata.org/pandas-docs/version/0.23.4/generated/pandas.DataFrame.from_csv.html)

but we're supposed to use `pd.DataFrame.from_dict` and not `pd.from_dict`

docs/data-hub/dataframes.md
sg-s: i agree. i've disabled auto-saving by default. users should be explicit if they want this behavior.

src/utils.py
sg-s: fixed

docs/ref/data-hub/types.md
sg-s: fixed

src/utils.py
sg-s: fixed

src/data_hub/dataframe.py
sg-s: fixed

src/data_hub/dataframe.py
sg-s: fixed

src/data_hub/api.py
sg-s: there's a lot of redundancy here. i'll clean up in a future pr

docs/data-hub/dataframes.md
sg-s: fixed

docs/data-hub/dataframes.md
sg-s: > because this is essential to understanding how this works.

i agree. i think of admonitions more as callouts than sidetracks. i'm using it to draw attention to this

docs/data-hub/dataframes.md
sg-s: done

docs/data-hub/dataframes.md
sg-s: fixed

docs/data-hub/dataframes.md
sg-s: fixed

docs/data-hub/dataframes.md
sg-s: fixed

docs/data-hub/dataframes.md
sg-s: fixed

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,130 @@
+# Using Deep Origin DataFrames
+
+This page describes how to use Deep Origin DataFrames, which are the primary object you will use to interact with databases on Deep Origin. This page will cover:
```
jonrkarr, line 3: Should "which are the primary object" be "which are the primary objects" (object plural) or "which is the primary class" (class singular)?

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,130 @@
...
+    ```
+
+In either case, when we view the DataFrame once more using `df`, we see the updated data, together with a warning telling us that we have local changes that haven't been written back to Deep Origin.
+
+
+![DataFrame with warning telling us about local changes that have not been written back to Deep Origin](../images/dataframe-1.png)
+
+## Write data back to Deep Origin
+
+### Using the `to_deeporigin` method
```
jonrkarr, line 70: Either at the beginning or end of this section, I think it would be helpful to outline what `to_deeporigin` can and cannot do:
- Update values of existing cells
  - Non-files
  - Files
- Update names of existing columns
- Update types of existing columns
- Create new rows
- Create new columns
- Remove rows
- Remove columns
- Create database

This could simply be a callout with a link to the "limitations" section below

My aim is simply to provide the user clarify so they know what to expect; we don't need to support all of the above.

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+!!! danger "Use `auto_sync` with caution"
+    Turning on `auto_sync` on dataframes can be dangerous. 
+    - Changes made to the local database are written to a Deep Origin database automatically, and no confirmation is asked for. 
+    - This can cause data loss.
+    - Every change made to the database is written immediately, so modifying the local dataframe multiple times leads to multiple writes to a Deep Origin database.
+
+## Limitations
+
```
jonrkarr, line 122: Before stating what this doesn't do, I think it would be instructive to add 1 more sentence that says what `to_deeporigin` does do -- update the values of existing cells.

docs/data-hub/dataframes.md
```diff
@@ -0,0 +1,130 @@
...
+# ✔︎ Wrote 9 rows in y^2 to Deep Origin database.
+```
+
+!!! danger "Use `auto_sync` with caution"
+    Turning on `auto_sync` on dataframes can be dangerous. 
+    - Changes made to the local database are written to a Deep Origin database automatically, and no confirmation is asked for. 
+    - This can cause data loss.
+    - Every change made to the database is written immediately, so modifying the local dataframe multiple times leads to multiple writes to a Deep Origin database.
+
+## Limitations
```
jonrkarr, line 121: I think this should be at the third level (`###`)

docs/data-hub/dataframes.md
jonrkarr, line 3: Should "which are the primary object" be "which are the primary objects" (object plural) or "which is the primary class" (class singular)?

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,219 @@
...
+    DatabaseReturnType,
+    IDFormat,
+    construct_resource_url,
+)
+
+
+class DataFrame(pd.DataFrame):
+    """A subclass of pandas DataFrame that allows for automatic updates to Deep Origin databases. This can be used as a drop-in replacement for a pandas DataFrame, and should support all methods a pandas DataFrame supports.
+
+    The primary method of creating an object of this type is to use the [api.get_dataframe][src.data_hub.api.get_dataframe] function.
```
jonrkarr, line 23: Should we update this to `from_deeporigin`?

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,219 @@
...
+        return_type: DatabaseReturnType = "dataframe",
+        client=None,
+    ):
+        """Create a local Deep Origin DataFrame from a Deep Origin database.
+
+        Args:
+        database_id (str): The ID of the Deep Origin database.
+        use_file_names (bool, optional): Whether to use the file names in the Deep Origin database. Defaults to True.
+        reference_format (IDFormat, optional): The format of the IDs in the Deep Origin database. Defaults to "human-id".
+        return_type (DatabaseReturnType, optional): The type of return value. Defaults to "dataframe".
```
jonrkarr, line 150: I think this needs to be nested

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,219 @@
...
+    def to_deeporigin(
+        self,
+        *,
+        columns: Optional[list] = None,
+        rows: Optional[list] = None,
+    ):
+        """Write data in dataframe to Deep Origin
+
+        !!! tip "Deep Origin DataFrames automatically synchronize"
+            Typically, you do not need to manually synchronize. If the `auto_sync` attribute of the dataframe is set to `True`, the dataframe will automatically synchronize when changes are made to the dataframe.
```
jonrkarr, line 171: With the change to auto-sync default=False, I suggest rephrasing to 

> To automatically save changes to local DataFrames to Deep Origin databases, set the `auto_sync` attribute of the dataframe `True`.

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+        Args:
+            columns (list, optional): The columns of the dataframe to update. Defaults to None.
```
jonrkarr, line 175: > Defaults to None.

I understand that the value is `None`. From a user's standpoint, should be clarify that the default is to update all columns? Add "When None, all modified columns are updated."?

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,219 @@
+"""
+This module defines a class called DataFrame that is a drop-in
+replacement for a pandas DataFrame, but also allows automatic
```
jonrkarr, line 3: I suggest removing "automatic" since its no longer the default behavior

src/data_hub/dataframe.py
```diff
@@ -0,0 +1,219 @@
...
+from deeporigin.data_hub import api
+from deeporigin.utils import (
+    DatabaseReturnType,
+    IDFormat,
+    construct_resource_url,
+)
+
+
+class DataFrame(pd.DataFrame):
+    """A subclass of pandas DataFrame that allows for automatic updates to Deep Origin databases. This can be used as a drop-in replacement for a pandas DataFrame, and should support all methods a pandas DataFrame supports.
```
jonrkarr, line 21: I suggest removing "automatic" since its no longer the default behavior

jonrkarr (approved):
Left several minor suggests for clarifying the documentation

docs/data-hub/dataframes.md
sg-s: fixed

docs/data-hub/dataframes.md
sg-s: made this an admonition

docs/data-hub/dataframes.md
sg-s: fixed

src/data_hub/dataframe.py
sg-s: fixed

src/data_hub/dataframe.py
sg-s: fixed

src/data_hub/dataframe.py
sg-s: fixed

src/data_hub/dataframe.py
sg-s: fixed

src/data_hub/dataframe.py
sg-s: fixed

src/data_hub/dataframe.py
sg-s: fixed

PR #74, opened 2024-08-29
fix: number of small bugfixes with dataframes

sg-s:
## changes

- [x] fixed a bug where link was broken for some dataframes
- [x] better representation of dataframe headers
- [x] fixed a bug where int columns were not typed correctly

### DIFF
--- a/docs/data-hub/dataframes.md
+++ b/docs/data-hub/dataframes.md
@@ -15,7 +15,7 @@ This page describes how to a Deep Origin DataFrame, which are the primary object
 Create a DataFrame using:
 
 ```python
-from deeporigin import DataFrame
+from deeporigin.data_hub.dataframe import DataFrame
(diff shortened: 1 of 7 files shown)


docs/data-hub/dataframes.md
```diff
@@ -15,7 +15,7 @@ This page describes how to a Deep Origin DataFrame, which are the primary object
 Create a DataFrame using:
 
 ```python
-from deeporigin import DataFrame
+from deeporigin.data_hub.dataframe import DataFrame
```
kennyjwilli, line 18: Does the AI assistant need to update?

kennyjwilli (approved):

docs/data-hub/dataframes.md
sg-s: it's not required. it can still use the old, explicit API:

```
api.get_dataframe
```

PR #75, opened 2024-08-30
feat: much faster file name resolution using ListFiles

sg-s:
## problem description

`api.get_dataframe` (and therefore `DataFrame.from_deeporigin`) can take extremely long because:

1. 2 calls to `list_database_rows` and `describe_row` take place in series when they could take place in parallel (this is a problem of 700ms vs 300ms)
2. (more serious problem). there is a place where N `DescriebFile` calls are made in series, where N is the number of files in the database. this means that the time it takes to fetch data grows with the number of files linearly, and the time for each file resolution is ~350ms. 

## proposed solution in this PR

using a new `ListFiles` endpoint that allows us to resolve all file IDs and names in a single API call (which typically takes 300ms no matter how many files)

### screenshot

from cold start, this is how long it takes to display a dataframe. in the old version, this took ~10s

Screenshot 2024-08-30 at 8 52 19 AM

## rejected alternative

using async methods. this approach was tried in this PR: https://github.com/deeporiginbio/deeporigin-client/pull/73 

it was rejected because it's much slower, and complicates testing dramatically


src/data_hub/api.py
```diff
@@ -904,65 +936,15 @@ def get_dataframe(
         return renamed_data
 
 
-@beartype
-def _parse_column_value(
```
sg-s, line 908: this wasn't being used anywhere. so deleted it.

src/data_hub/api.py
```diff
@@ -877,7 +910,6 @@ def get_dataframe(
         from deeporigin.data_hub.dataframe import DataFrame
 
         df = DataFrame(data)
-        df.attrs["file_ids"] = list(set(file_ids))
```
sg-s, line 880: also unused.

src/data_hub/api.py
```diff
@@ -817,25 +825,29 @@ def get_dataframe(
...
-    ]
-
     # figure out the column names and ID of the database
     db_row = _api.describe_row(
         row_id=database_id,
         client=client,
     )
-    assert (
-        db_row.type == "database"
-    ), f"Expected database_id: {database_id} to resolve to a database, but instead, it resolved to a {db_row.type}"
```
sg-s, line 838: we shouldn't have an assert here. we should raise the correct exception

src/data_hub/api.py
```diff
@@ -817,25 +825,29 @@ def get_dataframe(
...
     rows = _api.list_database_rows(
         database_row_id=database_id,
         client=client,
     )
 
-    # filter out template rows
-    rows = [
-        row for row in rows if not (hasattr(row, "is_template") and row.is_template)
-    ]
-
```
sg-s, line 830: this has moved a few lines down

src/data_hub/api.py
```diff
@@ -796,6 +796,14 @@ def download_database(
...
 
 
+def _replace_with_mapper(item, mapper: dict):
+    """utility function to replace items in a nested list with values generated by a mapper (a dict)."""
+
+    if isinstance(item, list):
+        return [_replace_with_mapper(sub_item, mapper) for sub_item in item]
+    return mapper.get(item, item)
+
+
```
sg-s, line 806: this arcane function is to deal with replacing file IDs in columns that can contain one, none, or many files

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "tabulate",
     "filetype",
     "httpx",
-    "deeporigin-data-sdk==0.1.0a7",
+    "deeporigin-data-sdk==0.1.0a8",
```
sg-s, line 30: this new version has support for the new argument in ListFiles

src/data_hub/_api.py
```diff
@@ -132,6 +135,7 @@ def dynamic_function(
                 tokens["access"] = auth.refresh_tokens(tokens["refresh"])
                 auth.cache_tokens(tokens)
                 client.token = tokens["access"]
+                method = _get_method(client, method_path)
```
jonrkarr, line 138: Thought: Is this necessary? Does `_get_method` directly use `client.token`?

jonrkarr (approved):
Looks good. Nice job identifying a bottleneck and finding a way to speed it up. 

Using async to execute many queries in parallel is a good thought. In some cases, we can get even higher performance with specialized bulk or composite methods.

src/data_hub/_api.py
sg-s: i **think** it's needed (not entirely sure). 

in any case, this is simply a wrapper around `getattr` so i don't think it's doing anything __bad__ or slow. 

i think it's needed because the new method belongs to the client object with the correct token (i think???) i may have to trace the logic step by step

PR #76, opened 2024-09-05
fix: better testing for config

sg-s:


.github/workflows/main.yml
```diff
@@ -70,18 +70,12 @@ jobs:
 
       - run: echo '${{ steps.cached_python_setup.outputs.cache-hit }}'
 
-      - run: 'mkdir ~/.deeporigin/'
-        shell: bash
-
```
sg-s, line 75: we don't need to do this because we now assume this folder doesn't exist

tests/test_cli_config.py
```diff
@@ -0,0 +1,52 @@
...
+        # get a list of all files
+        files = api.list_files()
+        if len(files) > 0:
+            data["file"] = files[0].file
+
+    # tests run on yield
+    yield data
+
+    # teardown tasks, if any
+
```
sg-s, line 44: this machinery will be needed later on for better testing

tests/test_cli_data.py
```diff
@@ -129,20 +128,6 @@ def test_show_db(config, json_option):
...
-    stderr = io.StringIO()
-
-    with redirect_stdout(stdout), redirect_stderr(stderr):
-        with cli.App(argv=argv) as app:
-            app.client = client
-            app.run()
-
-    return stdout.getvalue().strip()
-
-
```
sg-s, line 145: this has been moved to an utils folder so we can reuse it

jonrkarr (approved):

PR #83, opened 2024-09-30
fix: more bugfixes & improvements

sg-s:
## changes

- [x] wrapped all pretty printing code for data frames in a try so that this never breaks important code 
- [x] overrides for head and tail methods
- [x] better visibility in dark mode for dataframe headers
- [ ] ability to retrieve notebook entries 
- [x] ability to add a new column in a dataframe
- [x] caching of user names for faster displays

## technical discussion 

previously this behavior existed, where printing a dataframe worked, but using df.head() would trigger a spurious warning about dataframes containing modified data

dataframe-warnings 9 07 43 AM

this was because this was technically correct -- pandas df.head() created a shallow copy of a dataframe, which was technically a new object, and because the data in the new dataframe was not the same, the warning was shown.

however, this is useless and misleading from an end user persp. so we override these methods.

### DIFF
--- a/src/data_hub/dataframe.py
+++ b/src/data_hub/dataframe.py
@@ -165,12 +165,13 @@ class DataFrame(pd.DataFrame):
     def __repr__(self):
         """method override to customize printing in an interactive session"""
 
+        df_representation = super().__repr__()
         try:
             header = f'{self.attrs["metadata"]["hid"]}\n'
-            df_representation = super().__repr__()
-            return header + df_representation
         except Exception:
-            return super().__repr__()
+            header = ""
+
+        return header + df_representation
 
     @classmethod
     def from_deeporigin(
--- a/src/data_hub/dataframe.py
+++ b/src/data_hub/dataframe.py
@@ -41,7 +41,7 @@ class DataFrame(pd.DataFrame):
 
             return self.obj._get_value(*key)
 
-        def __setitem__(self, key, value):
+        def __setitem__(self, key, value) -> None:
             """intercept for the set operation""" ""
 
             old_value = self.obj._get_value(*key)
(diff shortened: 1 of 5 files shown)


src/data_hub/dataframe.py
```diff
@@ -81,66 +85,92 @@ def __setitem__(self, key, value):
...
 
-        header = f'{self.attrs["metadata"]["hid"]}\n'
-        df_representation = super().__repr__()
-        return header + df_representation
+        try:
+            header = f'{self.attrs["metadata"]["hid"]}\n'
+            df_representation = super().__repr__()
+            return header + df_representation
+        except Exception:
+            return super().__repr__()
```
kennyjwilli, line 173: Why would this throw?

kennyjwilli: Could this be slightly refactored to minimize the scope of the try? 

```python
df_representation = super().__repr__()
try:
    header = f'{self.attrs["metadata"]["hid"]}\n'
    return header + df_representation
except Exception:
    return df_representation
```

src/data_hub/dataframe.py
```diff
@@ -228,3 +268,57 @@ def _constructor(self):
...
         return DataFrame
+
+
+@beartype
+def _infer_column_type(column: pd.Series):
+    """utility function to infer type of data in a pandas column"""
+
+    non_null_values = column.dropna()
+
+    if non_null_values.empty:
```
kennyjwilli, line 279: Is it possible to defer the creation of the column until inference could at least make a reasonable guess? If it's empty, then the default could very likely be wrong.

kennyjwilli (approved):
nice 😎

src/data_hub/dataframe.py
sg-s: > Why would this throw?

if somehow `attrs` didn't contain the right fields

src/data_hub/dataframe.py
sg-s: column is created after inference. (we can't create a column without knowing what type...)

the failure mode of the inference code is to return `string`

src/data_hub/dataframe.py
sg-s: fixed

PR #84, opened 2024-10-01
feat: support for passing sha256 checksum to upload_file

sg-s:


src/data_hub/api.py
```diff
@@ -1,7 +1,6 @@
 """The `deeporigin.data_hub.api` module contains high-level functions for
 interacting with your Deep Origin data hub."""
 
-import mimetypes
```
sg-s, line 4: moved this into the only function that uses this

src/platform/api.py
```diff
@@ -24,14 +26,8 @@ def _make_request(endpoint: str) -> dict:
         "cache-control": "no-cache",
     }
 
-    env = get_value()["env"]
-    if env == "prod":
-        url = f"https://os.deeporigin.io/api{endpoint}"
-    else:
-        f"https://{env}.deeporigin.io/api{endpoint}"
-
```
sg-s, line 32: we use this over and over. so we're pulling this out

src/utils.py
```diff
@@ -275,10 +276,61 @@ def _get_method(obj, method_path):
...
+def _ensure_do_folder() -> None:
     """makes sure that ~/.deeporigin exists"""
 
     deeporigin_path = Path.home() / ".deeporigin"
 
     if not deeporigin_path.exists():
         deeporigin_path.mkdir(parents=True)
+
+
+def _download_nucleus_api_spec():
```
sg-s, line 288: this helps us download the spec for later

src/utils.py
```diff
@@ -275,10 +276,61 @@ def _get_method(obj, method_path):
...
+    """downloads the data hub API spec and saves to disk"""
+
+    _ensure_do_folder()
+
+    deeporigin_path = Path.home() / ".deeporigin"
+
+    url = urljoin(_get_domain_name(), "nucleus-api/api/openapi.json")
```
kennyjwilli, line 295: `nucleus-api/` -> `data-hub`

src/utils.py
```diff
@@ -275,10 +276,61 @@ def _get_method(obj, method_path):
...
+
+    print(url)
```
kennyjwilli, line 297: Intentional?

src/utils.py
kennyjwilli: Curious - why do we need it downloaded to disk?

kennyjwilli (approved):
Awesome, thanks for adding so quick

src/utils.py
sg-s: fixed

src/utils.py
sg-s: fixed

src/utils.py
sg-s: solely for me to update the SDK later on!

PR #85, opened 2024-10-01
refactor: split up utils into different modules

sg-s:
## changes

- [x] large, overdue refactor splitting up utils into more sensible chunks
- [x] fixes a bug where this wouldn't work on the lambda because we don't have write access to ~/.deeporigin


src/utils/config.py
```diff
@@ -0,0 +1,59 @@
...
+    Returns:
+        str: URL for the resource
+    """
+
+    env = get_value()["env"]
+    org = get_value()["organization_id"]
+    if env == "prod":
+        url = f"https://os.deeporigin.io/org/{org}/data/{row_type}/{name}"
+    else:
+        url = f"https://os.{env}.deeporigin.io/org/{org}/data/{row_type}/{name}"
```
kennyjwilli, line 34: It'd be nice to unify this code.

src/utils/config.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def _get_domain_name() -> str:
+    """utility function to get domain name based on env"""
+
+    env = get_value()["env"]
+    if env == "prod":
+        return "https://os.deeporigin.io"
+    else:
+        return f"https://os.{env}.deeporigin.io"
```
kennyjwilli, line 59: Ideally this also gets deduplicated with the above code.

src/utils/constants.py
```diff
@@ -0,0 +1,53 @@
+"""this module contains constants used in the rest of this library"""
+
+from dataclasses import dataclass
+
+from beartype.typing import Literal
+
+ObjectType = Literal["row", "database", "workspace"]
```
kennyjwilli, line 7: Really such a shame there's no way to pull this from the openapi spec.

src/utils/core.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@beartype
+def _ensure_do_folder() -> Path:
+    """makes sure that the deeporigin scratch folder exists
+
+    If it doesn't exist, the folder is created. The location of this folder is typically in the home directory, but may be elsewhere if running in AWS Lambda"""
+
+    if in_aws_lambda():
+        deeporigin_path = Path("/tmp") / ".deeporigin"
```
kennyjwilli, line 155: `.deeporigin` to top-level constant?

kennyjwilli (approved):
Much more well-organized!

src/utils/config.py
sg-s: oooof good catch! this needs to use the method i wrote in the last PR

sg-s: fixed

src/utils/config.py
sg-s: fixed

src/utils/constants.py
sg-s: this will be addressed in a future PR

PR #86, opened 2024-10-02
refactor: dropped support for graphql secrets and variables

sg-s:
## changes

- [x] ~~added crypt + jwt packages~~
- [x] ~~added function to decode token~~
- [x] removed graphql request
- [x] added rest API platform secret method
- [x] patched construction of pydantic types for variables and secrets
- [x] updated codeowners
- [x] updated docs to take into account refactoring of utils
- [x] refactored utils
- [x] disabled some tests


src/platform/api.py
```diff
@@ -1,17 +1,17 @@
...
 import requests
 from beartype import beartype
 from deeporigin import auth
 from deeporigin.config import get_value
+from deeporigin.utils.core import _ensure_do_folder
 from deeporigin.utils.network import _get_domain_name
 
 
-def _make_request(endpoint: str) -> dict:
+def _make_get_request(endpoint: str) -> dict:
```
jonrkarr, line 14: Minor: this specifically implements `get` (not also `post`, `delete`, etc.) requests. The previous name reflected this.

src/platform/api.py
```diff
@@ -87,3 +89,13 @@ def get_last_edited_user_name(row):
     else:
         user_id = row.edited_by_user_drn.split(":")[-1]
     return get_user_name(user_id)
+
+
+@beartype
+def get_secrets() -> list[dict]:
```
jonrkarr, line 95: In most places we call this "variables and secrets" because some are encrypted and some are not. I suggest aligning the terminology in the method name.

src/platform/api.py
sg-s: @jonrkarr so the current name is better, right?

jonrkarr: You're right. I mixed up the previous/new.

src/auth.py
```diff
@@ -23,8 +23,8 @@
 @beartype
 def tokens_exist() -> bool:
     """Check if the cached API tokens exist"""
-    config = get_config()
-    return os.path.isfile(config.api_tokens_filename)
+
+    return os.path.isfile(_get_api_tokens_filepath())
```
sg-s, line 28: we now use _get_api_tokens_filepath as the single source of truth

src/auth.py
```diff
@@ -80,26 +80,16 @@ def cache_tokens(tokens: dict) -> None:
...
 
 
-def read_cached_tokens() -> dict[str, str]:
-    """Read cached API tokens"""
-    config = get_config()
-    with open(config.api_tokens_filename, "r") as file:
-        tokens = json.load(file)
-    return tokens
-
-
```
sg-s, line 97: moved out of auth

src/cli/config.py
```diff
@@ -45,8 +49,6 @@ def show(self):
 
         data = get_value()
 
-        data.pop("list_workstation_variables_query_template", None)
-
```
sg-s, line 49: this field has been eliminated

src/cli/config.py
```diff
@@ -120,7 +122,7 @@ def load(self):
 
         name = self.app.pargs.profile
 
-        file_to_load = os.path.expanduser(f"~/.deeporigin/{name}.yml")
```
sg-s, line 123: removed references to ~/.deeporigin

src/config/__init__.py
```diff
@@ -22,16 +22,13 @@
     "env": confuse.String(),
     "api_endpoint": confuse.Optional(confuse.String()),
     "nucleus_api_route": confuse.String(),
-    "graphql_api_route": confuse.String(),
```
sg-s, line 25: no more graphql api route

src/config/default.yml
```diff
@@ -3,64 +3,13 @@ organization_id: null
...
 nucleus_api_route: nucleus-api/api/
-graphql_api_route: api/graphql/
 auth_domain: https://formicbio.us.auth0.com
 auth_device_code_endpoint: oauth/device/code/
 auth_token_endpoint: oauth/token/
 auth_audience: https://os.deeporigin.io/api
 auth_grant_type: urn:ietf:params:oauth:grant-type:device_code
 auth_client_id: m3iyUcrANcIap2ogzWKpnYxCNujOrW3s
 auth_client_secret: cQcZclTqMHMuovyXV-DD15tEiL-KH_2XD36vsppULRBuq7AjwyI4dh5ag11O_K1S
-list_workstation_variables_query_template: |
```
sg-s, line 14: @jonrkarr i think we don't need this any more

src/data_hub/api.py
```diff
@@ -302,7 +302,7 @@ def upload_file(
         "Connection": "keep-alive",
         "Content-Length": str(content_length),
         "Content-Type": content_type,
-        "x-amz-checksum-sha256": hash,
+        # "x-amz-checksum-sha256": hash, # TODO: add checksum
```
sg-s, line 305: @kennyjwilli commenting this out to work with what's on prod

src/data_hub/api.py
```diff
@@ -316,7 +316,10 @@ def upload_file(
         )
 
         if put_response.status_code != 200:
-            raise DeepOriginException(message="File could not be uploaded.")
+            raise DeepOriginException(
+                title=f"❌ File could not be uploaded! [{put_response.status_code}]",
+                message=f"The response was: {put_response.text}",
+            )
```
sg-s, line 322: a better error message. hopefully we'll never need it

src/platform/api.py
```diff
@@ -26,8 +26,10 @@ def _make_request(endpoint: str) -> dict:
         "cache-control": "no-cache",
     }
 
+    url = urljoin(_get_domain_name(), f"api/{endpoint}")
+
     response = requests.get(
-        urljoin(_get_domain_name(), "api", endpoint),
```
sg-s, line 30: this was a bug -- urljoin only works with 2 args

src/utils/core.py
```diff
@@ -160,3 +160,30 @@ def _ensure_do_folder() -> Path:
...
+
+# @beartype
+# def decode_access_token() -> dict:
+#     """decode token and extract info"""
+
+#     import jwt
+
+#     tokens = read_cached_tokens()
+
+#     return jwt.decode(tokens["access"], options={"verify_signature": False})
```
sg-s, line 173: this is commented out for now. it'll come back in a future pr

src/utils/core.py
```diff
@@ -160,3 +160,30 @@ def _ensure_do_folder() -> Path:
...
+    return _ensure_do_folder() / "api_tokens"
+
+
+@beartype
+def read_cached_tokens() -> dict:
+    """Read cached API tokens"""
+
+    with open(_get_api_tokens_filepath(), "r") as file:
+        tokens = json.load(file)
+    return tokens
```
sg-s, line 189: this has been moved from auth

src/variables/core.py
```diff
@@ -274,6 +263,8 @@ def deserialize_variable(serialized: dict) -> Variable:
     Returns:
         :obj:`Variable`: variable
     """
+    serialized = serialized.copy()
```
sg-s, line 266: needed to insert this because this function was mutating input args

tests/test_variables.py
```diff
@@ -1673,104 +1660,6 @@ def test_get_auto_install_variables_cronjob_id(self):
             )
         )
 
-    def test_get_remove_do_api_tokens(self):
-        config.get_value.cache_clear()
-        feature_flags.get_value.cache_clear()
-        auth.get_tokens.cache_clear()
-
-        _, self.api_tokens_filename = tempfile.mkstemp(suffix=".yml")
```
sg-s, line 1681: this test has been deleted because we have removed the ability to configure the location of api_tokens via a config file.

this will be centrally managed and, in a later PR, there will be support for mulitple of these files

tests/test_variables.py
```diff
@@ -2437,9 +2327,7 @@ def test_install_secret_file_value(self):
         user_home_dirname = tempfile.mkdtemp()
 
         expected_filenames = []
-        self.assertEqual(
-            sorted(expected_filenames), sorted(os.listdir(user_home_dirname))
-        )
+        self.assertEqual(set(expected_filenames), set(os.listdir(user_home_dirname)))
```
sg-s, line 2330: switched from sorted lists to sets because sets give us cleaner error messages

src/platform/api.py
sg-s: fixed

kennyjwilli (approved):

PR #88, opened 2024-10-04
feat: added ability to download files in parallel

sg-s:
## changes

- [x] ability to download mulitple files in parallel using ~~asyncio~~ concurrent.futures
- [x] docs improvements
- [x] added changelog to docs
- [x] added docs on how to download files
- [x] added methods to download files from the CLI
- [x] added docs on platform API features 
- [x] ~~download_database should use download_files, not download_file~~ next sprint
- [x] download_files should allow users to download to a specific folder -- no need to specify all names
- [x] download_files should work with no args -- and it will download all files
- [x] dropped support for python 3.9

## technical discussion

there are two ways we can go about downloading files. one is to use asyncio and the other is to use concurrent.futures

i initially implemented everything asyncio, but decided to switch to concurrent.futures because

- mixing async with sync code leads to a lot of boilerplate and repeated code
- asyncio code doesn't work natively in jupyter notebooks. we have to use a 3rd party package called nested_asyncio to get it to work, which is more overhead
- technically asyncio should be lighter and faster, but i didn't see this in practice 
- using asyncio also requires us to use the low level async client, which leads to more doubling of work and code
- using the low-level asyncio client makes testing more complex, because we can't hot swap clients 

### rejected asyncio implementation:

```python
async def _create_file_download_urls_async(file_ids: list[str]) -> list[str]:
    """async method to create file download URLs for a list of files.

    Do not use this method. This is called internally by the
    `download_files` method."""

    async_client = _api._get_default_client(use_async=True)

    tasks = []
    for file_id in file_ids:
        tasks.append(
            _api.create_file_download_url(file_id=file_id, client=async_client)
        )
    data = await asyncio.gather(*tasks)
    urls = [item.data.download_url for item in data]
    return urls

```

and 

```python
@beartype
def download_files(
    file_ids: Optional[list[str]] = None,
    names: Optional[list[str]] = None,
):
    """download multiple files in parallel using asyncio

    If you want to download a single file, use download_file as it has lower overhead.

    Args:
        file_ids: IDs of the files on Deep Origin
        names: Names of the files. Optional. If None, names will be retrieved from Deep Origin

    Returns:
        None

    """

    # we need nest_asynio to allow this work
    # in a jupyter kernel
    import nest_asyncio

    nest_asyncio.apply()

    if names is None:
        # names not provided, determine names from Deep Origin
        files = list_files(file_ids=file_ids)
        names = [item.file.name for item in files]

    # create presigned URLs for all files in parallel
    urls = asyncio.run(_create_file_download_urls_async(file_ids))

    # download files in parallel
    asyncio.run(_download_files_async(urls, names))

    return urls
```

and 

```python
async def _download_async(session, url, save_path) -> None:
    """Downloads a single file asynchronously and saves it to the specified path.

    Do not use this. Use the synchronous wrapper function
    download_files instead."""

    async with session.get(url) as response:
        with open(save_path, "wb") as file:
            async for chunk in response.content.iter_chunked(8192):
                if chunk:  # Filter out keep-alive chunks
                    file.write(chunk)


async def _download_files_async(urls: list[str], save_paths: list[str]) -> None:
    """Downloads multiple files asynchronously.

    Do not use this. Use the synchronous wrapper function
    download_files instead."""

    async with aiohttp.ClientSession() as session:
        tasks = []
        for url, save_path in zip(urls, save_paths):
            tasks.append(_download_async(session, url, save_path))

        await asyncio.gather(*tasks)
````

### CODE REVIEW

File: .github/workflows/main.yml [hunk 1]
```diff
@@ -55,7 +55,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
```
sg-s, line 58: dropped support for py3.9 because it's old and we would like to use new type annotation syntax

docs/data-hub/dataframes.md
```diff
@@ -136,4 +136,4 @@ df
 
 ## Reference
 
-Read more about the `to_deeporigin` method [here](../ref/data-hub/types.md#src.data_hub.dataframe.DataFrame.sync).
```
sg-s, line 139: this was an error, not sure how docs built before

icon.svg
sg-s, line 1: not sure why we had an icon -- wasn't being used anywhere

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
     "humanize",
     "packaging",
     "diskcache",
+    "tqdm",
```
sg-s, line 34: added tqdm for progress bars in downloads. we'll use this elsewhere too in later prs

src/cli/config.py
```diff
@@ -122,7 +122,7 @@ def load(self):
 
         name = self.app.pargs.profile
 
-        file_to_load = _ensure_do_folder / f"{name}.yml"
```
sg-s, line 125: this was a bug

src/config/default.yml
```diff
@@ -2,7 +2,7 @@ env: prod
 organization_id: null
 bench_id: null
 api_endpoint: https://os.deeporigin.io
-nucleus_api_route: nucleus-api/api/
+nucleus_api_route: data-hub/api/
```
sg-s, line 5: @kennyjwilli the change has been made

src/data_hub/_api.py
```diff
@@ -88,11 +89,20 @@ def _get_default_client(
...
+                token=access_token,
+                org_id=org_id,
+                base_url=base_url,
+            )
+        else:
+            client = DeeporiginData(
+                token=access_token,
+                org_id=org_id,
+                base_url=base_url,
+            )
```
sg-s, line 105: i've implemented support for async, but we're not using it anywhere. in future we might

src/data_hub/api.py
```diff
@@ -993,6 +997,65 @@ def get_dataframe(
         return renamed_data
 
 
+@beartype
+@ensure_client
+def download_files(
+    files: Optional[list[ListFilesResponse] | ListFilesResponse] = None,
```
sg-s, line 1003: better stronger typing for args

src/utils/core.py
```diff
@@ -26,6 +26,14 @@ def _get_method(obj, method_path):
...
 
 
+def humanize_file_size(file_size: int):
+    """humanizes a file size in bytes"""
+    for unit_prefix in ["", "K", "M", "G", "T", "P", "E", "Z", "Y"]:
+        if file_size < 1024.0:
+            return f"{file_size:.2f} {unit_prefix}B"
+        file_size /= 1024.0
+
+
```
sg-s, line 36: this pretty prints file sizes

src/utils/network.py
```diff
@@ -11,7 +12,7 @@
 
 
 @beartype
-def download_sync(url: str, save_path: str) -> None:
+def download_sync(url: str, save_path: str | Path) -> None:
```
sg-s, line 15: also supports pathlib Paths. in future we want to move away from str representation of paths.

src/config/default.yml
kennyjwilli: I assume no issues?

src/platform/api.py
```diff
@@ -33,7 +34,15 @@ def _make_get_request(endpoint: str) -> dict:
...
-    return response.json()
+    if response.status_code != 200:
+        raise DeepOriginException(message=response.text, fix=url)
+
+    data = response.json()
+    if "data" in data.keys():
+        return data["data"]
+
+    else:
+        return data
```
kennyjwilli, line 45: Convenient, yes, but it will likely bite later 😉

kennyjwilli (approved):
Nice!

PR #89, opened 2024-10-08
feat: added ability to compute hash on file upload

sg-s:


src/data_hub/api.py
```diff
@@ -265,6 +265,8 @@ def download_file(
 def upload_file(
     file_path: str,
     client=None,
+    *,
+    compute_hash: bool = False,
```
kennyjwilli, line 269: Why defaulting off?

kennyjwilli (approved):

jonrkarr:
Looks good, except for the comment about the default. Don't we want it to be on for AWS?

src/data_hub/api.py
sg-s: ok, turning on by default!

sg-s: fixed

PR #92, opened 2024-10-15
feat: ability to decode tokens

sg-s:


src/platform/api.py
```diff
@@ -108,3 +111,41 @@ def get_variables_and_secrets() -> list[dict]:
...
 
     return response["data"]
+
+
+@functools.cache
+@beartype
+def get_public_keys() -> list[dict]:
+    """get public keys from public endpoint"""
+
+    jwks_url = urljoin(get_value()["auth_domain"], ".well-known/jwks.json")
```
kennyjwilli, line 121: IMO, you should push `get_value()["auth_domain"]` as far up the call stack as possible. Not blocking.

src/platform/api.py
```diff
@@ -108,3 +111,41 @@ def get_variables_and_secrets() -> list[dict]:
...
+    data = requests.get(jwks_url).json()
+    return data["keys"]
+
+
+@beartype
+def decode_access_token() -> dict:
+    """decode access token into human readable data"""
+
+    tokens = auth.get_tokens()
```
kennyjwilli, line 130: Ditto. Idea being -- imagine writing tests for these functions, as is. You'll have to mock `auth` and `get_value`. Alternatively, if you had pure versions of these functions, you could just pass values directly in a test.

kennyjwilli (approved):

src/platform/api.py
sg-s: i get you now. `get_value` is expected to work in tests too -- we set up a dummy config file

PR #93, opened 2024-10-15
fix: suppress message about new version on lambda

sg-s:
## changes

previously, using the AI assistant would surface this message, which cannot be acted upon because the user can't upgrade the client

Screenshot 2024-10-15 at 11 56 06 AM

### DIFF
--- a/src/utils/network.py
+++ b/src/utils/network.py
@@ -1,14 +1,32 @@
 """this module contains network related utility functions"""
 
+import functools
 import json
 from pathlib import Path
 from urllib.parse import parse_qs, urljoin, urlparse
 
 import requests
 from beartype import beartype
+from deeporigin import __version__
 from deeporigin.exceptions import DeepOriginException
 from deeporigin.utils.config import _get_domain_name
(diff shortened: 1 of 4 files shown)


src/utils/network.py
```diff
@@ -1,14 +1,32 @@
...
 from beartype import beartype
+from deeporigin import __version__
 from deeporigin.exceptions import DeepOriginException
 from deeporigin.utils.config import _get_domain_name
-from deeporigin.utils.core import _ensure_do_folder
+from deeporigin.utils.core import _ensure_do_folder, in_aws_lambda
+from packaging.version import Version
+
+
+@functools.cache
```
sg-s, line 17: we use functools.cache to make sure this is shown only once per session

src/utils/network.py
```diff
@@ -1,14 +1,32 @@
...
+def check_for_updates():
+    """check if a new version is available. If so, print an message"""
+    if not in_aws_lambda():
+        try:
```
sg-s, line 21: everything is inside a try because this should never block

kennyjwilli (approved):

PR #94, opened 2024-10-17
feat: ability to delete terminated workstations

sg-s:
## changes

- [x] ability to delete terminated workstations
- [x] ability to delete any workstation
- [x] refactor: single method to make requests with any verb
- [x] expanded testing to py 3.13

### DIFF
--- a/src/platform/api.py
+++ b/src/platform/api.py
@@ -1,5 +1,6 @@
-"""module to interact with the platform API"""
+"""experimental module to interact with the platform API"""
 
+import concurrent.futures
 import functools
 from typing import Optional
 from urllib.parse import urljoin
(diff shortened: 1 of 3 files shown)


src/platform/api.py
```diff
@@ -35,13 +46,25 @@ def _make_get_request(endpoint: str) -> dict:
...
+    if verb == "GET":
+        response = requests.get(
+            url,
+            headers=headers,
+        )
+    elif verb == "POST":
+        response = requests.post(
+            url,
+            headers=headers,
+        )
```
kennyjwilli, line 58: Perhaps nicer if you use [`requests.request`](https://requests.readthedocs.io/en/latest/api/#requests.request)?

src/platform/api.py
```diff
@@ -51,29 +74,78 @@ def _make_get_request(endpoint: str) -> dict:
         return data
 
 
+@beartype
+def delete_workstation(workstation: str):
+    """delete a workstation"""
+
+    return _make_request(
+        f"computebenches/{_get_org_id()}/{workstation}:delete", verb="POST"
```
kennyjwilli, line 82: Probably not in scope here, but client pattern? 😁

src/platform/api.py
```diff
@@ -51,29 +74,78 @@ def _make_get_request(endpoint: str) -> dict:
...
+        if ws["attributes"]["status"] == "TERMINATED"
+    ]
+
+    if len(names) == 0:
+        print("No workstations to delete")
+        return
+
+    org_id = _get_org_id()
+
+    with concurrent.futures.ThreadPoolExecutor() as executor:
```
kennyjwilli, line 104: Can you bound the size of the thread pool? Or is it by default?

kennyjwilli (approved):
General feedback, non-blocking. LGTM.

src/platform/api.py
sg-s: lol much nicer :)

src/platform/api.py
sg-s: the default is ncores

src/platform/api.py
sg-s: yes....also needs to use the autogenerated low level sdk

src/platform/api.py
sg-s: fixed

PR #96, opened 2024-10-23
fix: better typing for dataframes

sg-s:
## changes

- [x] support for expressions in dataframes -- results shown directly
- [x] ~~writing to expressions/formulas from a dataframe should be disallowed~~ pending fix to formulas
- [x] support for rendering Person names in dataframes
- [ ] support for writing Person info in dataframes and pushing it
- [x] support for references
- [x] ~~support for reference look-ups~~ for future PR
- [x] tests on a kitchen sink DB
- [x] modifying a subset of a column should not write the entire column back, but only the updated cells
- [x] references to rows resolved and row names rendered 
- [x] set_data_in_cells shouldn't call descibe_row if column metadata is passed

### writing a single cell back to the DB

Screenshot 2024-10-23 at 11 35 49 AM

### writing whole columns back to the DB

Screenshot 2024-10-23 at 11 36 57 AM

### people are now resolved!

Screenshot 2024-10-23 at 11 37 44 AM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,8 +68,6 @@ jobs:
           cache: 'pip'
           cache-dependency-path: pyproject.toml
 
-      - run: echo '${{ steps.cached_python_setup.outputs.cache-hit }}'
-
       - name: "run tests without ~/.deeporigin/"
         run: |
           pip install --upgrade pip
--- a/src/cli/data.py
+++ b/src/cli/data.py
@@ -343,6 +343,7 @@ class DataController(cement.Controller):
             data = api.get_dataframe(
                 self.app.pargs.object_id,
                 return_type="dict",
+                reference_format="system-id",
                 client=self._get_client(),
             )
 
--- a/src/data_hub/api.py
+++ b/src/data_hub/api.py
@@ -1002,7 +1002,7 @@ def get_dataframe(
         df.attrs["last_updated_row"] = find_last_updated_row(rows)
 
         df._deep_origin_out_of_sync = False
-        df._modified_columns = set()
+        df._modified_columns = dict()
         return df
 
     else:
--- a/src/data_hub/api.py
+++ b/src/data_hub/api.py
@@ -18,6 +18,7 @@ from deeporigin.data_hub import _api
 # functions in _api are available in api (this module)
 from deeporigin.data_hub._api import *  # noqa: F403
 from deeporigin.exceptions import DeepOriginException
+from deeporigin.platform.api import get_user_name
 from deeporigin.utils.constants import (
     PREFIXES,
     Cardinality,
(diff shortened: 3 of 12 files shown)


.github/workflows/main.yml
```diff
@@ -68,8 +68,6 @@ jobs:
           cache: 'pip'
           cache-dependency-path: pyproject.toml
 
-      - run: echo '${{ steps.cached_python_setup.outputs.cache-hit }}'
```
sg-s, line 71: this was a debug statement

src/cli/data.py
```diff
@@ -343,6 +343,7 @@ def show(self):
             data = api.get_dataframe(
                 self.app.pargs.object_id,
                 return_type="dict",
+                reference_format="system-id",
```
sg-s, line 346: human-id needs more complex changes to the mock client

src/data_hub/api.py
```diff
@@ -546,19 +547,23 @@ def set_data_in_cells(
...
-    )
+
+    if columns is None:
+        # we need to get the columns and their types
+        response = _api.describe_row(
+            row_id=database_id,
+            client=client,
+        )
+
+        columns = response.cols
```
sg-s, line 563: this allows us to pass in columns, and bypass a call to DescribeRow

src/data_hub/api.py
```diff
@@ -961,6 +969,21 @@ def get_dataframe(
...
+        for conversion in conversions:
+            ref_id_mapper[conversion.id] = conversion.hid
+
+        for column in columns:
+            if column["type"] == "reference":
+                inputs = data[column["id"]]
+
+                data[column["id"]] = [
+                    _replace_with_mapper(item, ref_id_mapper) for item in inputs
+                ]
```
sg-s, line 985: map and replace system ID -> human ID

src/data_hub/api.py
```diff
@@ -1089,17 +1122,24 @@ def _row_to_dict(row, *, file_ids: list):
             continue
         if field.value is None:
             value = None
-        elif field.type in ["float", "int", "boolean"]:
```
sg-s, line 1092: this was wrong

src/data_hub/api.py
```diff
@@ -1089,17 +1122,24 @@ def _row_to_dict(row, *, file_ids: list):
...
         elif field.type == "file":
             value = field.value.file_ids
             file_ids.extend(value)
+        elif field.type == "expression":
+            value = field.value.result
+        elif field.type == "user":
+            user_ids = field.value.user_drns
+
+            # convert to names using the platform API
+            value = [get_user_name(user_id) for user_id in user_ids]
```
sg-s, line 1142: support for expressions and users

src/data_hub/api.py
```diff
@@ -1089,17 +1122,24 @@ def _row_to_dict(row, *, file_ids: list):
...
 
         elif field.type == "reference":
             value = field.value.row_ids
-
+            reference_ids.extend(value)
         elif field.type == "file":
             value = field.value.file_ids
             file_ids.extend(value)
+        elif field.type == "expression":
+            value = field.value.result
```
kennyjwilli, line 1137: Technically the `type` of an expression column is the return type specified in the column definition.

src/data_hub/dataframe.py
```diff
@@ -190,48 +194,39 @@ def from_deeporigin(
...
 
+            rows = columns[column]
+
             column_metadata = [
                 col for col in self.attrs["metadata"]["cols"] if col["name"] == column
             ]
 
             if len(column_metadata) == 0:
+                # new column to add
+                print("Creating new column...")
```
kennyjwilli, line 229: Intentional?

kennyjwilli (approved):

src/data_hub/dataframe.py
sg-s: in this case, yes!

src/data_hub/api.py
sg-s: we'll have to dig deeper into this. expressions not well supported in this PR because there are issues with them in the kitchen sink DB

PR #97, opened 2024-10-23
feat: slicing a dataframe using `loc` preserves attributes

sg-s:
## changes

- [x] loc indexer preservers attrs
- [x] add a new prop: `_allow_new_rows`
- [x] df indexed using loc should disallow adding new rows via loc

## in a future PR

- ability to write data back to DB when adding new rows
- support for appending a pandas dataframe

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,7 @@ endif
 	interrogate -c pyproject.toml -v . -f 100 && \
 	python3 -m coverage run -m pytest -x -n $(n_workers) --failed-first -k $(chosen_tests) --client $(client) && \
 	python3 -m coverage html && \
-	deactivate && \
-	open htmlcov/index.html; \
+	deactivate
 
 # set up jupyter dev kernel
 jupyter:
--- a/src/data_hub/dataframe.py
+++ b/src/data_hub/dataframe.py
@@ -31,6 +31,24 @@ class DataFrame(pd.DataFrame):
     _modified_columns: dict = dict()
     """if data is modified in a dataframe, and auto_sync is False, this list will contain the columns that have been modified so that the Deep Origin database can be updated. If an empty list, the Deep Origin database will not be updated, and the dataframe matches the Deep Origin database at the time of creation."""
 
+    @property
+    def loc(self):
+        class _LocIndexer:
+            def __init__(self, df):
+                self.df = df
+
+            def __getitem__(self, key):
+                # first call the superclass method
+                df = super(DataFrame, self.df).loc[key]
+
+                # inherit attributes
+                df.attrs = self.df.attrs
+
+                return df
+
+        # Return the custom _LocIndexer instance
+        return _LocIndexer(self)
+
     class AtIndexer:
         """this class override is used to intercept calls to at indexer of a pandas dataframe"""
 
--- a/src/data_hub/dataframe.py
+++ b/src/data_hub/dataframe.py
@@ -58,6 +58,8 @@ class DataFrame(pd.DataFrame):
                 return df
 
             def __setitem__(self, key, value):
+                """callback for adding a new row, typically"""
+
                 if not self.df._allow_adding_rows:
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -27,8 +27,7 @@ endif
 	interrogate -c pyproject.toml -v . -f 100 && \
 	python3 -m coverage run -m pytest -x -n $(n_workers) --failed-first -k $(chosen_tests) --client $(client) && \
 	python3 -m coverage html && \
-	deactivate && \
-	open htmlcov/index.html; \
+	deactivate
```
sg-s, line 30: removed this because this was becoming annoying

src/data_hub/dataframe.py
```diff
@@ -31,6 +34,46 @@ class DataFrame(pd.DataFrame):
...
+    _allow_adding_rows: bool = True
+    """If `True`, new rows can be added to the dataframe. If `False`, new rows cannot be added to the dataframe."""
+
+    @property
+    def loc(self):
+        class _LocIndexer:
+            def __init__(self, df):
+                self.df = df
+
+            def __getitem__(self, key):
```
sg-s, line 46: this is a LocIndexer, which overrides the superclass class attribute. allows us to define what happens when people slice and index into their dataframe

src/data_hub/dataframe.py
```diff
@@ -31,6 +34,46 @@ class DataFrame(pd.DataFrame):
     _modified_columns: dict = dict()
     """if data is modified in a dataframe, and auto_sync is False, this list will contain the columns that have been modified so that the Deep Origin database can be updated. If an empty list, the Deep Origin database will not be updated, and the dataframe matches the Deep Origin database at the time of creation."""
 
+    _allow_adding_rows: bool = True
```
sg-s, line 37: the reason for disallowing adding new rows is as follows:

let's say you have a DB with 10 rows. you then slice into the first 2. at that point, those other 8 rows are still there in the DB -- so you should not be allowed to add a row that interferes with those 8 rows

tests/test_dataframe.py
```diff
@@ -1,23 +1,43 @@
...
 import pytest
 from deeporigin.data_hub import api
 from deeporigin.data_hub.dataframe import DataFrame
 
-from tests.utils import clean_up_test_objects, minimal_config
+from tests.utils import clean_up_test_objects
+
+TEST_PREFIX = "tc-"
 
-TEST_PREFIX = "tc-x1zrncgf6F-"
```
sg-s, line 10: these unique IDs are now dynamically generated, so that tests can be run in parallel

tests/test_dataframe.py
```diff
@@ -1,23 +1,43 @@
...
-TEST_DB_NAME = TEST_PREFIX + "db"
 COLUMNS = ["float", "integer"]
 
+SKIP_MSG = "Skipping this test because we can't test this well using a mocked client"
+
 
-@pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="function", autouse=True)
```
sg-s, line 20: so that every test calls this fixture separately, and it is re-run each time

tests/test_dataframe.py
```diff
@@ -64,27 +84,100 @@ def config(pytestconfig):
...
 
+        data["db-name"] = TEST_DB_NAME
+
     # tests run on yield
     yield data
 
     # teardown code
     if pytestconfig.getoption("client") != "mock":
-        clean_up_test_objects(TEST_PREFIX)
+        clean_up_test_objects(TEST_PREFIX + salt)
```
sg-s, line 96: clean up resources created just by this test

src/data_hub/dataframe.py
kennyjwilli: I don't follow -- how would adding a new row interfere with the other 8?

kennyjwilli (approved):

PR #102, opened 2024-10-28
fix: insertion of series data into dataframe

sg-s:
## changes

inline

### DIFF
--- a/src/data_hub/dataframe.py
+++ b/src/data_hub/dataframe.py
@@ -19,6 +19,11 @@ from deeporigin.utils.network import check_for_updates
 check_for_updates()
 
 
+try:
+    from icecream import ic
+except ImportError:  # Graceful fallback if IceCream isn't installed.
+    ic = lambda *a: None if not a else (a[0] if len(a) == 1 else a)  # noqa
+
 __NO_NEW_ROWS_MSG__ = "Adding rows is not allowed, because this dataframe corresponds to a subset of the rows in the corresponding database."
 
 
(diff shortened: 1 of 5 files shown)


src/data_hub/dataframe.py
```diff
@@ -138,8 +143,12 @@ def __setitem__(self, key, value):
...
             self.to_deeporigin()
         else:
-            # an empty set means "all "
-            self._modified_columns[key] = set()
+            if isinstance(key, str):
+                # an empty set means "all "
+                self._modified_columns[key] = set()
+            elif isinstance(key, list):
+                for item in key:
+                    self._modified_columns[item] = set()
```
sg-s, line 151: this fix is for the weird code that the AI assistant wrote where series were inserted into dataframes for every row

sg-s (commented):
the AI assistant wrote some correct but unexpected and idiosyncratic code, so need this patch to make sure it works as intended

src/data_hub/dataframe.py
```diff
@@ -19,6 +19,11 @@
 check_for_updates()
 
 
+try:
+    from icecream import ic
+except ImportError:  # Graceful fallback if IceCream isn't installed.
+    ic = lambda *a: None if not a else (a[0] if len(a) == 1 else a)  # noqa
```
kennyjwilli, line 25: Could we pull this into a utility file since it seems to be reused?

kennyjwilli (approved):

src/data_hub/dataframe.py
sg-s: decided to remove it because it's trivial to re-add when debugging, and this will never be neeeded in prod

PR #107, opened 2024-10-30
fix: annotation improvements

sg-s:
## changes

- [x] fixed a bug where swapping axes works
- [x] support for coloring points when we label them
- [x] removed support for choosing the select column at runtime
- [x] support for reading in existing labels from the database 
- [x] reading in labels from a database should show colors correctly

### DIFF
--- a/src/data_hub/axes_callback.js
+++ b/src/data_hub/axes_callback.js
@@ -2,10 +2,16 @@
 // this is called every time the user makes a selection
 // on what column of the dataframe to plot onto what axes
 
+
+console.log(df);
+console.log(df["ID"]);
+
 const x_data = df[x_select.value];
 const y_data = df[y_select.value];
 const size_data = df[size_select.value];
-const id_data = df[ID.value];
+const id_data = df["ID"];
+
+
 
 // Normalize size data for better visualization (e.g., map values to a range)
 const min_size = Math.min(...size_data);
--- a/src/data_hub/axes_callback.js
+++ b/src/data_hub/axes_callback.js
@@ -3,9 +3,6 @@
 // on what column of the dataframe to plot onto what axes
 
 
-console.log(df);
-console.log(df["ID"]);
-
 const x_data = df[x_select.value];
 const y_data = df[y_select.value];
 const size_data = df[size_select.value];
--- a/src/data_hub/axes_callback.js
+++ b/src/data_hub/axes_callback.js
@@ -9,8 +9,6 @@ console.log(df["ID"]);
 const x_data = df[x_select.value];
 const y_data = df[y_select.value];
 const size_data = df[size_select.value];
-const id_data = df["ID"];
-
 
 
 // Normalize size data for better visualization (e.g., map values to a range)
@@ -23,7 +21,9 @@ const sizes = size_data.map(value => {
 });
 
 // Update the data source
-scatter_source.data = { 'x': x_data, 'y': y_data, 'size': sizes, 'id': id_data };
+scatter_source.data.x = x_data;
+scatter_source.data.y = y_data;
+scatter_source.data.size = sizes;
 
 // Update the axis labels
 x_axis.axis_label = x_select.value;
--- a/src/data_hub/button_callback.js
+++ b/src/data_hub/button_callback.js
@@ -34,5 +34,8 @@ if (window.deeporigin) {
         ]
     }));
 
-    deeporigin.dataHub.primaryDatabase.editRows({ changes: updateChanges });
+    deeporigin.dataHub.primaryDatabase.editRows({
+        checkPreviousValue: false, 
+        changes: updateChanges
+    });
 }
--- a/src/data_hub/button_callback.js
+++ b/src/data_hub/button_callback.js
@@ -1,24 +1,31 @@
-// Callback when a user presses the button to update labels in the scatter plot Bokeh figure
+// Callback to update labels in the scatter plot Bokeh figure
 
 const selectedData = lasso_selection_source.data;
 const label = label_select.value;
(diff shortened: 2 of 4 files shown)


src/data_hub/axes_callback.js
```diff
@@ -17,7 +18,9 @@ const sizes = size_data.map(value => {
 });
 
 // Update the data source
-scatter_source.data = { 'x': x_data, 'y': y_data, 'size': sizes, 'id': id_data };
```
sg-s, line 20: no idea why i was updating the whole object

src/data_hub/axes_callback.js
```diff
@@ -2,10 +2,11 @@
 // this is called every time the user makes a selection
 // on what column of the dataframe to plot onto what axes
 
+
 const x_data = df[x_select.value];
 const y_data = df[y_select.value];
 const size_data = df[size_select.value];
-const id_data = df[ID.value];
```
sg-s, line 8: axes callback shouldn't need or mess with ID at all

src/data_hub/plotting.py
```diff
@@ -24,6 +26,8 @@ def scatter(
     y: Optional[str] = None,
     size: Optional[str] = None,
     hover_callback_code: Optional[str] = None,
+    button_callback_code: Optional[str] = None,
+    label_column: Optional[str] = None,
```
sg-s, line 30: these 2 JS files are the only ones that interact with the deeporigin JS API, so allowing us to override it at execution time for faster debug cycles. 

this will eventually be removed

src/data_hub/plotting.py
```diff
@@ -109,8 +146,9 @@ def scatter(
         y="y",
         source=scatter_source,
         size="size",
-        color="blue",
+        color="colors",
```
sg-s, line 149: this is highly confusing, but this now refers to a column called `colors` in the CDS, not the literal string `blue` as it previously did

src/data_hub/plotting.py
```diff
@@ -149,44 +187,14 @@ def scatter(
...
-            // Update options of the second select based on unique values of the selected column
-            const column_data = label_source.data[column];
-            const unique_values = Array.from(new Set(column_data));
-            
-            label_select.options = unique_values;
-            label_select.value = unique_values[0];  // Set default to the first unique value
-        """,
-        )
-
-        label_column_select.js_on_change("value", label_select_callback)
```
sg-s, line 188: removed the ability to choose, in JS, the column to write labels to

src/data_hub/plotting.py
```diff
@@ -209,6 +217,9 @@ def scatter(
     if hover_callback_code is None:
         hover_callback_code = js_code["hover_callback"]
 
+    if button_callback_code is None:
+        button_callback_code = js_code["button_callback"]
+
```
sg-s, line 222: added ability to override JS code for button callback

src/data_hub/button_callback.js
```diff
@@ -1,24 +1,37 @@
...
+// get the points within the lasso selection
 const selectedData = lasso_selection_source.data;
+
+// get the label we need to assign
 const label = label_select.value;
-const labelColumn = label_column_select.value;
 
-console.log("Will write to these rows:", selectedData.ids);
-console.log("Will write this label:", label);
-console.log("Will write to this column:", labelColumn);
```
sg-s, line 9: removed debug code

src/data_hub/hover_callback.js
```diff
@@ -13,6 +13,28 @@ if (cb_data.index && cb_data.index.indices && cb_data.index.indices.length > 0)
...
+        var value = data[col][chosenIndex];
+
+        if (typeof value === 'number' && !Number.isInteger(value)) {
+            value = value.toFixed(3);
+        }
+        values.push(value);
+    }
+    table_source.data = { 'Column Name': columnNames, 'Value': values };
+    table_source.change.emit();
+
```
sg-s, line 37: this updates the table that corresponds to each row on hover. note that we hide some columns

src/data_hub/plotting.py
```diff
@@ -23,7 +27,8 @@ def scatter(
     x: Optional[str] = None,
     y: Optional[str] = None,
     size: Optional[str] = None,
-    hover_callback_code: Optional[str] = None,
+    label_column: Optional[str] = None,
```
sg-s, line 30: now we need to supply the column that contains labels at runtime

src/data_hub/plotting.py
```diff
@@ -38,22 +43,58 @@ def scatter(
...
     """
 
+    # constants for this plot
     figure_width = 500
     select_width = int(figure_width * 0.3)
-    js_code = _read_js_code()
+    if js_code is None:
+        js_code = _read_js_code()
+    default_label = "None"
+    default_color = "blue"
```
sg-s, line 52: moved some more constants here

src/data_hub/plotting.py
```diff
@@ -38,22 +43,58 @@ def scatter(
...
+                f"The column {label_column} has too many ({len(labels)}) options configured, and therefore cannot be used to label data."
+            )
+
+        # make color map
+        cat10_colors = Category10
+        cat10_colors[1] = cat10_colors[3][:1]
+        cat10_colors[2] = cat10_colors[3][:2]
+        cat10_colors = cat10_colors[len(labels)]
+        color_map = {label: color for label, color in zip(labels, cat10_colors)}
+        color_map[default_label] = default_color
```
sg-s, line 91: this inscrutable code makes a colormap, which works for anything from 1-10 colors

kennyjwilli (approved):

PR #114, opened 2024-11-18
Refactor of test stack

sg-s:
## changes

see comments inline

## problem description

our underlying stainless-generated code returned responses as objects. this was problematic because:

- sometimes these pydantic models would fail to `model_dump`
- this introduced a dependence on these models
- tests became more complex because mock client needed to recreate all these objects

## solution

- switch to only dealing with dicts
- overhaul the mock client to directly read from a stash of responses from the live client


Makefile
```diff
@@ -10,12 +10,7 @@ repo=$(shell basename $(CURDIR))
 # a live instance 
 client="mock"
 chosen_tests=""
-
-lint:
-	@source $(CURDIR)/venv/bin/activate && \
-		ruff format && \
-		ruff check --select I && \
-		deactivate
```
sg-s, line 18: don't need this

Makefile
```diff
@@ -25,7 +20,7 @@ else
 endif 
 	@source $(CURDIR)/venv/bin/activate && \
 	interrogate -c pyproject.toml -v . -f 100 && \
-	python3 -m coverage run -m pytest -x -n $(n_workers) --failed-first -k $(chosen_tests) --client $(client) && \
+	python3 -m coverage run -m pytest -x -n $(n_workers) --failed-first -k $(chosen_tests) --client $(client) --responses $(responses) && \
 	python3 -m coverage html && \
```
sg-s, line 24: adding --responses to determine whether we should stash responses or not

pyproject.toml
```diff
@@ -27,13 +27,14 @@ dependencies = [
     "tabulate",
     "filetype",
     "httpx",
-    "deeporigin-data-sdk==0.1.0a12",
+    "deeporigin-data-sdk==0.1.0a21",
```
sg-s, line 30: this is the thing that has triggered all of this. we were stuck on v12 for a while

src/cli/data.py
```diff
@@ -220,12 +218,12 @@ def list(self):
             pdata = dict(Name=[], Type=[], ID=[])
 
             for item in rows:
-                pdata["Name"].append(item.name)
+                pdata["Name"].append(getattr(item, "name", None))
+
```
sg-s, line 222: this construction (you'll see this everywhere) works even when there is no attribute named item

src/data_hub/_api.py
```diff
@@ -48,7 +54,7 @@ def _get_client_methods():
     # the only reason we're creating this client is to
     # extract methods from it. So no need to
     # authenticate
-    client = DeeporiginData(token="", org_id="")
+    client = DeeporiginData(token="", org_id="").with_raw_response
```
sg-s, line 57: this change tells the client to not construct pydantic models and simply return the JSON the server gives us

src/data_hub/_api.py
```diff
@@ -182,18 +183,38 @@ def dynamic_function(
...
+            )
+            stashed_data = PersistentDict(stash_loc)
+            key = json.dumps(kwargs, sort_keys=True)
+
+            if isinstance(response, list):
+                stashed_response = response[:10]
+            else:
+                stashed_response = response
+
+            stashed_data[key] = stashed_response
```
sg-s, line 215: if _stash, then we save the responses in a JSON file on disk. this will then be consumed by the MockClient, which allows us to exactly reproduce the state of a live instance using a local mock, without needing any HTTP traffic

tests/test_data_hub.py
```diff
@@ -166,17 +170,18 @@ def test_list_files_assigned(config):  # noqa: F811
...
 
 def test_describe_database_stats(config):  # noqa: F811
     stats = api.describe_database_stats(
         database_id=config["databases"][0],
         client=config["client"],
+        _stash=config["stash"],
     )
 
-    assert stats.row_count >= 0, "Expected a positive number of rows"
+    assert stats.rowCount >= 0, "Expected a positive number of rows"
```
sg-s, line 185: more of this stuff: camelCase <- snake_case

tests/test_data_hub.py
```diff
@@ -375,29 +405,7 @@ def test_download_database(config):  # noqa: F811
...
-    )
-
-    file_name = db_id + ".csv"
-    assert os.path.exists(file_name), f"Expected to find a CSV file called {file_name}"
-    os.remove(file_name)
-
-    api.download(
-        f"{PREFIXES.DO}{db_id}",
-        destination=os.getcwd(),
-        client=config["client"],
```
sg-s, line 400: this is essentially a duplicated test and introduces a race condition when run in parallel (as is the case when we run against a live instance). so removing

tests/test_cli_config.py
```diff
@@ -1,50 +1,12 @@
...
-        files = api.list_files()
-        if len(files) > 0:
-            data["file"] = files[0].file
-
-    # tests run on yield
-    yield data
-
-    # teardown tasks, if any
-
-
```
sg-s, line 46: we don't need any of these. refactored to use a common config

tests/mock_client.py
sg-s, line 1: this is where the refactor really shines. all of this code is gone. MockClient simply reads from JSON files and responds with previously stashed data

src/data_hub/_api.py
```diff
@@ -182,18 +180,38 @@ def dynamic_function(
...
-            print("Response headers:")
-            _print_dict(dict(response.headers), json=False)
-            print(f"Response = {response.http_response}")
+        if not isinstance(response, dict):
             response = response.json()
 
-        if hasattr(response, "data"):
-            return response.data
+        if "data" in response.keys():
+            response = response["data"]
```
kennyjwilli, line 187: Worth thinking about how to deal with paginated requests, which will include paginated information at the top-level of `response`. Food for thought.

src/data_hub/_api.py
```diff
@@ -182,18 +180,38 @@ def dynamic_function(
...
+            if isinstance(response, list):
+                response = [Box(item) for item in response]
+            else:
+                response = Box(response)
         else:
-            # this should never happen
-            return response
+            response = Box(response)
+
+        if _stash:
```
kennyjwilli, line 195: Why?

src/data_hub/api.py
```diff
@@ -890,15 +974,25 @@ def get_dataframe(
...
+            client=client,
+            _stash=_stash,
+        )
+    else:
+        rows = _api.list_database_rows(
+            database_row_id=database_id,
+            client=client,
+            _stash=_stash,
+            filter=filter,
+        )
```
kennyjwilli, line 989: Is this really the way to pass an optional param? Can't `list_database_rows` handle an filter value of None?

src/data_hub/filters.py
kennyjwilli, line 1: Why is this necessary? Validation only? Could we instead move the validation to the operation invocation level? e.g., instead of validating some subset of request data, validate the whole thing prior to sending the request. Put another way, why are these special cased?

tests/fixtures/responses/create_database.json
kennyjwilli, line 1: Curious - how do you plan to keep these updated as the API adds new fields to responses?

kennyjwilli (approved):
This looks like a huge win wrt simplification. Hopefully you got all the snake -> camel cases! Let's test 🚢

tests/fixtures/responses/create_database.json
sg-s: i'm going to set up a github action to run periodically and check for a change in spec, and regenerate those responses.

src/data_hub/filters.py
sg-s: that is a good question. right now it's validation only but as you guessed this is half baked

src/data_hub/api.py
sg-s: i get a bad request error

```
BadRequestError: Error code: 400 - {'errors': [{'title': 'invalid request', 'status': 400, 'meta': {'issues': [{'code': 'invalid_union', 'unionErrors': [{'issues': [{'code': 'invalid_union', 'unionErrors': [{'issues': [{'code': 'invalid_type', 'expected': 'object', 'received': 'null', 'path': ['filter'], 'message': 'Expected object, received null'}], 'name': 'ZodError'}, {'issues': [{'code': 'invalid_type', 'expected': 'object', 'received': 'null', 'path': ['filter'], 'message': 'Expected object, received null'}], 'name': 'ZodError'}, {'issues': [{'code': 'invalid_type', 'expected': 'object', 'received': 'null', 'path': ['filter'], 'message': 'Expected object, received null'}], 'name': 'ZodError'}, {'issues': [{'code': 'invalid_type', 'expected': 'object', 'received': 'null', 'path': ['filter'], 'message': 'Expected object, received null'}], 'name': 'ZodError'}], 'path': ['filter'], 'message': 'Invalid input'}], 'name': 'ZodError'}, {'issues': [{'code': 'invalid_type', 'expected': 'object', 'received': 'null', 'path': ['filter'], 'message': 'Expected object, received null'}], 'name': 'ZodError'}, {'issues': [{'code': 'invalid_type', 'expected': 'object', 'received': 'null', 'path': ['filter'], 'message': 'Expected object, received null'}], 'name': 'ZodError'}, {'issues': [{'code': 'invalid_type', 'expected': 'object', 'received': 'null', 'path': ['filter'], 'message': 'Expected object, received null'}], 'name': 'ZodError'}], 'path': ['filter'], 'message': 'Invalid input'}]}}]}
```

kennyjwilli: When you pass `None`, instead of converting to `null`, can you not set the key?

PR #115, opened 2024-11-21
Add rows

sg-s:
## changes

- [x] loc indexer should disallow adding new rows to a DF
- [x] ability to modify existing rows using loc
- [x] from_deeporigin now works with completely empty DBs
- [x] from_deeporigin now works with DBs with no rows (but some columns)
- [x] tests: ability to modifiy 1 row using .loc
- [x] tests: ability to modify many rows using .loc
- [x] tests: prevention of adding new rows to a DF
- [x] support for appending a square chunk (similar to a DB) directly to a DB. combination of create new rows + write data
- [x] support for writing 1 new row to a DB with all columns with data
- [x] support for writing many rows + many columns
- [x] documentation explaining how to add new rows to a DB


docs/how-to/data-hub/add-data.md
```diff
@@ -0,0 +1,43 @@
...
+## Add fragments of new rows 
+
+Similarly, fragments of rows (subsets of columns) can be written to the database:
+
+```python
+data = dict(X=[10, 20]) # note Y is not specified
+api.add_database_rows(database_id="xy", data=data)
+```
+
+`add_database_rows` returns a list of the row IDs created during this process.
```
kennyjwilli, line 38: I'd find an example return value useful.

src/data_hub/api.py
```diff
@@ -358,11 +358,81 @@ def upload_file(
...
+    Args:
+        database_id: Human ID or System ID of the database
+        data: A dictionary where each key is a column name and each value is a list of values. All values should have the same length. Key names should match column names in the database.
+
+    Returns:
+        A list of row IDs
+
+    """
+    # check that dict has columns that make sense
+    db = _api.describe_row(
```
kennyjwilli, line 382: Switch to `DescribeDatabase`

src/data_hub/api.py
```diff
@@ -358,11 +358,81 @@ def upload_file(
...
+        _stash=_stash,
+    )
+
+    col_names = [col.name for col in db.cols]
+
+    for col in data.keys():
+        if col not in col_names:
+            raise DeepOriginException(
+                message=f"Column `{col}` does not exist in database `{database_id}`."
+            )
```
kennyjwilli, line 394: The server will also do this check, if you'd prefer to not do it here.

docs/how-to/data-hub/add-data.md
```diff
@@ -0,0 +1,43 @@
...
+
+![](../../images/df-xy.png)
+
+## Add new rows
+
+To add new rows to the underlying database, use the `add_databse_rows` function:
+
+```python
+data = dict(X=[1, 2], Y=[2, 3])
+api.add_database_rows(database_id="xy", data=data)
```
kennyjwilli, line 21: Do people tend to think of adding data by columns or rows? I would've leaned more towards the latter.

kennyjwilli (approved):
I'd lean more towards a row-centric add method, but if you think column centric is more pythonic, let's do that.

src/data_hub/api.py
sg-s: i'd rather catch errors early!

docs/how-to/data-hub/add-data.md
sg-s: fixed

src/data_hub/api.py
sg-s: fixed

src/data_hub/api.py
kennyjwilli: It is a possible race condition, though fairly unlikely.

kennyjwilli:
> > I'd lean more towards a row-centric add method, but if you think column centric is more pythonic, let's do that.
> 
> @kennyjwilli not sure what you mean? it kinda _is_ row centric in that you get to add rows.

You pass in columns first instead of rows. Personally, I find passing rows first much easier to work with.

PR #116, opened 2024-11-21
Support for platform API

sg-s:
## changes


Makefile
```diff
@@ -38,7 +38,7 @@ install:
 	@python3 -m venv venv
 	@source $(CURDIR)/venv/bin/activate && \
 		pip install --upgrade pip && \
-	    pip install -e .[lint,test,dev,docs,plots] && \
+	    pip install --no-cache-dir -e .[lint,test,dev,docs,plots] && \
 	    deactivate
```
sg-s, line 42: this forces us to not use the cache. this is useful to avoid errors when hot swapping low-level modules. this only affects me, users never see this

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
     "pyjwt",
     "cryptography",
     "python-box",
+    "do-sdk-platform==1.0.0",
```
sg-s, line 38: added auto-generated platform SDK

src/platform/utils.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Returns:
+        configured client
+    """
+
+    if configure:
+        configuration = do_sdk_platform.configuration.Configuration(
+            host=urljoin(get_value()["api_endpoint"], "/api"),
+            access_token=get_tokens()["access"],
+        )
```
sg-s, line 29: configure with token and URL, which comes from the main library

src/platform/utils.py
```diff
@@ -0,0 +1,106 @@
...
+
+        client = do_sdk_platform.ApiClient(configuration=configuration)
+    else:
+        client = do_sdk_platform.ApiClient()
```
sg-s, line 33: if configure = False, we don't need it configured, we just need that object to inspect and dynamically build functions

src/platform/utils.py
```diff
@@ -0,0 +1,106 @@
...
+    """utility function to get methods from the client that return raw responses from the server"""
+    methods = set(
+        [
+            attr
+            for attr in dir(client)
+            if callable(getattr(client, attr))
+            and not attr.startswith("_")
+            and "without_preload_content" in attr
+        ]
+    )
```
sg-s, line 51: we're getting all (public) methods that have `without_prelload_content` in them so that we can use raw responses. typically errors in the spec mean that we can't reliably construct objects from them

src/platform/workstations.py
```diff
@@ -0,0 +1,36 @@
...
+        this_module_name,
+        sanitized_method_name,
+        _create_function(
+            method_path=method,
+            api_name=API_NAME,
+        ),
+    )
+
+
+__all__ = list(methods)
```
kennyjwilli, line 36: Suggest pulling this out to reusable fn.

kennyjwilli (approved):

src/platform/workstations.py
sg-s: done!

PR #118, opened 2024-11-22
fix: better handling of text responses

sg-s:
## changes

small fix to handle text responses

### DIFF
--- a/src/data_hub/_api.py
+++ b/src/data_hub/_api.py
@@ -7,7 +7,6 @@ with these functions via api.py
 import inspect
 import json
 import sys
-from json import JSONDecodeError
 from pathlib import Path
 
 from beartype import beartype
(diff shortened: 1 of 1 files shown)


src/data_hub/_api.py
```diff
@@ -183,12 +184,17 @@ def dynamic_function(
             )
 
         if not isinstance(response, dict):
-            response = response.json()
+            try:
+                response = response.json()
+            except JSONDecodeError:
+                response = dict(data=response.text())
```
kennyjwilli, line 190: Fine for now, but a better solution would be to dispatch on the response `content-type` header.

kennyjwilli (approved):

src/data_hub/_api.py
sg-s: i should check for `'text/event-stream'`?

sg-s: looks like it's either `application/json` or `text/event-stream`?

sg-s: fixed

PR #120, opened 2024-11-25
fix: only patch the action

sg-s:
## what

tests against a live instance (that were configured to run daily) were broken because the access token kept going stale. 

this PR fixes this by manually refreshing before running tests. 

## proof 

https://github.com/deeporiginbio/deeporigin-client/actions/runs/12014629544

### DIFF
--- a/.github/workflows/live_tests.yaml
+++ b/.github/workflows/live_tests.yaml
@@ -9,11 +9,27 @@ on:
     - cron: "0 0 * * *"
 
 jobs:
-  write-tokens:
+  test-live-instance:
     runs-on: ubuntu-latest
-    outputs:
-      api_tokens_path: ~/.deeporigin/api_tokens
+    env:
+      # Secrets for API and organization
+      DEEP_ORIGIN_REFRESH_TOKEN: ${{ secrets.DEEP_ORIGIN_REFRESH_TOKEN }}
+      DEEP_ORIGIN_ACCESS_TOKEN: ${{ secrets.DEEP_ORIGIN_ACCESS_TOKEN }}
+      DEEP_ORIGIN_ORGANIZATION_ID: ${{ secrets.DEEP_ORIGIN_ORGANIZATION_ID }}
+      DEEP_ORIGIN_API_ENDPOINT: ${{ secrets.DEEP_ORIGIN_API_ENDPOINT }}
+      DEEP_ORIGIN_NUCLEUS_API_ROUTE: ${{ secrets.DEEP_ORIGIN_NUCLEUS_API_ROUTE }}
+
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python environment
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: 'pip'
+          cache-dependency-path: pyproject.toml
+
       - name: Create API Tokens File
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/live_tests.yaml
```diff
@@ -1,36 +1,52 @@
...
+  test-live-instance:
     runs-on: ubuntu-latest
     env:
+      # Secrets for API and organization
+      REFRESH_TOKEN: ${{ secrets.DEEP_ORIGIN_REFRESH_TOKEN }}
+      ACCESS_TOKEN: ${{ secrets.DEEP_ORIGIN_ACCESS_TOKEN }}
       DEEP_ORIGIN_ORGANIZATION_ID: ${{ secrets.DEEP_ORIGIN_ORGANIZATION_ID }}
       DEEP_ORIGIN_API_ENDPOINT: ${{ secrets.DEEP_ORIGIN_API_ENDPOINT }}
       DEEP_ORIGIN_NUCLEUS_API_ROUTE: ${{ secrets.DEEP_ORIGIN_NUCLEUS_API_ROUTE }}
-      DEEP_ORIGIN_ACCESS_TOKEN: ${{ secrets.DEEP_ORIGIN_ACCESS_TOKEN }}
```
sg-s, line 18: we can't create a env variable called this because then the client will use this. we don't want it to -- we want it to read from file

.github/workflows/live_tests.yaml
```diff
@@ -1,36 +1,52 @@
...
 
-      - name: "Running tests against a live instance..."
+      - name: Create API Tokens File
+        run: |
+          mkdir -p ~/.deeporigin
+          echo '{
+            "refresh": "'$REFRESH_TOKEN'",
+            "access": "'$ACCESS_TOKEN'"
+          }' > ~/.deeporigin/api_tokens
+
```
sg-s, line 40: here we're writing to disk, so that it behaves like it would on end users (and my) computer

.github/workflows/live_tests.yaml
```diff
@@ -1,36 +1,52 @@
...
+      - name: Install dependencies
         run: |
           pip install --upgrade pip
           pip install --upgrade setuptools
           pip install -e .[test]
-          make test-github-live
+
+      - name: Refresh API tokens using deeporigin.auth
+        run: |
+          python -c "from deeporigin import auth; auth.get_tokens(refresh=True)"
```
sg-s, line 49: here we're using the client to refresh the token (which will typically always be stale)

.github/workflows/live_tests.yaml
```diff
@@ -1,36 +1,52 @@
...
+    - cron: "0 0 * * *"
 
 jobs:
-  test-against-live-instance:
+  test-live-instance:
     runs-on: ubuntu-latest
     env:
+      # Secrets for API and organization
+      REFRESH_TOKEN: ${{ secrets.DEEP_ORIGIN_REFRESH_TOKEN }}
+      ACCESS_TOKEN: ${{ secrets.DEEP_ORIGIN_ACCESS_TOKEN }}
```
sg-s, line 17: it is important that these are not prepended with DEEP_ORIGIN

kennyjwilli: Perhaps leave a comment why, haha.

kennyjwilli (approved):

PR #121, opened 2024-11-25
fix: removed download_file

sg-s:
## changes

- removed `download_file`
- refactored rest of code to use `download_files`

### DIFF
--- a/src/data_hub/api.py
+++ b/src/data_hub/api.py
@@ -241,48 +241,6 @@ def list_rows(
 
 @beartype
 @ensure_client
-def download_file(
-    file_id: str,
-    *,
-    destination: str | Path = os.getcwd(),
-    client=None,
-    _stash: bool = False,
-) -> None:
-    """Download a file to a destination folder (workspace).
-
-    Download a file synchronously from Deep Origin
-    to folder on the local file system.
-
-    Args:
-        file_id: ID of the file on Deep Origin
-        destination: Path to the destination folder
-
-    """
-
-    if not os.path.isdir(destination):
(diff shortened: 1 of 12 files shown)


tests/fixtures/responses/describe_database_stats.json
```diff
@@ -1,5 +1,5 @@
 {
     "{\"database_id\": \"kitchen-sink\"}": {
-        "rowCount": 127
+        "rowCount": 0
```
kennyjwilli, line 3: This correct?

src/data_hub/api.py
```diff
@@ -239,48 +239,6 @@ def list_rows(
     )
 
 
-@beartype
-@ensure_client
-def download_file(
```
kennyjwilli, line 244: Something to think about: if an existing code block uses this function, the user will not be able to re-run the code block without first updating to `download_files`. This may be especially confusing if AI wrote the code.

kennyjwilli (approved):

tests/fixtures/responses/describe_database_stats.json
sg-s: that is weird, isn't it? i don't think it's used anywhere

src/data_hub/api.py
sg-s: hmm, yes, it's a breaking change for sure. good catch. i'll include a stub to use download_files under the hood

sg-s: actually changed my mind. the LLM can never generate this code because it doesn't know about this function

PR #129, opened 2024-12-09
fix: works with new auth_domain

sg-s:
## changes

- [x] auth_domain for nucleus now matches auth_domain for platform-api
- [x] fixed a bug where the default config was read, not the active one 
- [x] fixed an endpoint on platform API that has become a PATCH

### DIFF
--- a/src/auth.py
+++ b/src/auth.py
@@ -108,7 +108,6 @@ def authenticate() -> dict:
 
     # Get a link for the user to sign into the Deep Origin platform
     endpoint = urljoin(config.auth_domain, config.auth_device_code_endpoint)
-    print(endpoint)
     body = {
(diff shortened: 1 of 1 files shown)


src/platform/api.py
```diff
@@ -132,7 +132,7 @@ def resolve_user(user_id: str):
 def whoami() -> dict:
     """get details about currently signed in user"""
 
-    return _make_request("/users/me")
+    return _make_request("/users/me", verb="PATCH")
```
kennyjwilli, line 135: This seems sketchy

kennyjwilli (approved):

src/platform/api.py
sg-s: this will go away when we use the auto generated client

PR #130, opened 2024-12-12
Support for filtering rows in a database

sg-s:
## changes

- [x] ability to filter rows in a database and populate dataframe from that
- [x] ability to automatically translate column names to IDs, so that users can specify column names


docs/data-hub/dataframes.md
```diff
@@ -39,6 +39,23 @@ which should show you something like this:
...
+        filter_value=42,
+        operator="greaterThan",
+        filter_type="number",
+    )
+    df = DataFrame.from_deeporigin("your-db-id",   filter=filter)
+    df
+```
+
+
+
```
sg-s, line 58: we'll need more docs but this at least tells the user something

src/data_hub/api.py
```diff
@@ -991,20 +998,26 @@ def get_dataframe(
...
+        column_names = [col.name for col in db_row.cols]
+        if filter.column_id not in column_ids:
+            column_id = [col.id for col in db_row.cols if col.name == filter.column_id]
+            if len(column_id) == 1:
+                filter.column_id = column_id[0]
+            else:
+                raise DeepOriginException(
+                    f"Filter column with ID or name: {filter.column_id} not found in database {database_id}",
+                    fix=f"Valid column names are: {column_names}",
+                )
```
sg-s, line 1012: convert from column name to ID

src/data_hub/api.py
```diff
@@ -983,6 +983,13 @@ def get_dataframe(
 
     # TODO: list_database_rows and describe_row can be called in parallel
 
+    # figure out the column names and ID of the database
+    db_row = _api.describe_row(
```
sg-s, line 987: this needs to be converted to describe_database _row, will happen later

src/data_hub/filters.py
```diff
@@ -1,65 +1,43 @@
...
-
-
-    Args:
-        column_id (str): id of the column
-        value (float | int): value to filter
-        operator (str): operator to use.
-
-    Returns:
-        A dictionary containing a valid condition that can be used to filter database rows
-    """
```
sg-s, line 28: this is pointless boilerplate and i don't like depending on the low level SDK for these types

src/data_hub/filters.py
```diff
@@ -1,65 +1,43 @@
...
+    "contains",
+    "notContains",
+    "startsWith",
+    "endsWith",
+    "substructure",
+    "in",
+    "notIn",
+    "isNull",
+    "isNotNull",
+]
```
sg-s, line 25: in future this will have to be read from the api spec

docs/data-hub/dataframes.md
```diff
@@ -39,6 +39,23 @@ which should show you something like this:
     - Information about the last edit made to the database. 
 
 
+!!! tip "Fetching only some rows"
+    If you have a particularly large database, you may want to fetch only certain rows using a filter. For example, to fetch only rows where column `x` is greater than `42`:
+
+    ```python
+    from deeporigin.data_hub import filters
+    filter = filters.filter(
```
kennyjwilli, line 47: Why do you need to do `filters.filter`?

kennyjwilli (approved):

docs/data-hub/dataframes.md
sg-s: it's in its own module, which is weird. we'll fix when we have a nicer way of doing this

PR #133, opened 2024-12-27
Draco

aregsyan:


docs/tools/draco.md
```diff
@@ -0,0 +1,84 @@
+# Draco: a Chemical Data Extractor
+
+[Draco] is a tool which aims to extract chemical data (Molecules, Reactions, procedures, etc.) from documents. In its first version, Draco focuses at extracting molecules (fragments and complete structures) from PDF documents
```
sg-s, line 3: we have to say that this is decimer

docs/tools/draco.md
```diff
@@ -0,0 +1,84 @@
...
+
+Upload a PDF file containing chemical structures to extract to the Data Hub.
+
+## Parameters
+
+This section describes the parameters for a tool run, that are passed in the `start_run` function.
+
+### IUPAC name (OPTIONAL) - NOT YET AVAILABLE
+
+Use a model to predict the IUPAC name of the predicted structures
```
sg-s, line 17: let's remove this

docs/tools/draco.md
```diff
@@ -0,0 +1,84 @@
+# Draco: a Chemical Data Extractor
```
sg-s, line 1: are we OK using this name in publicly facing documentation? i thought this was an internal project name

docs/tools/draco.md
```diff
@@ -0,0 +1,84 @@
...
+
+### 2. Start a tool run on Deep Origin
+
+For this tool run, we will use the following parameters:
+
+```python
+parameters = {
+    "input_file": 'your_pdf_file',
+}
+```
```
sg-s, line 39: why is this here? why not in the function argument?

docs/tools/draco.md
```diff
@@ -0,0 +1,84 @@
...
+
+## RoadMap
+
+- Add Reaction extraction
+- Extraction from Text
+- TODO
+
+## TODO
+
+Show a template picture of the data hub
```
sg-s, line 84: remove

src/tools/draco.py
```diff
@@ -0,0 +1,74 @@
+"""module to run AutoDock Vina on Deep Origin"""
```
sg-s, line 1: fix

src/tools/draco.py
```diff
@@ -0,0 +1,74 @@
...
+    db = api.describe_database(database_id=database_id)
+    cols = db.cols
+
+    # resolve columns
+    output_column_id = _resolve_column_name(output_column_name, cols)
+    input_file_column_id = _resolve_column_name(input_file_column_name, cols)
+
+    if not row_id.startswith("_row"):
+        data = data = api.convert_id_format(hids=[row_id])
+        row_id = data[0].id
```
sg-s, line 45: we actually don't need this. new versions of mason can work with HIDs

sg-s (changes requested):

src/tools/draco.py
aregsyan: Actually it was a datahub functionality and it was working always. We only have issue with columns

src/tools/draco.py
```diff
@@ -0,0 +1,70 @@
...
+        data = api.convert_id_format(hids=[database_id])
+        database_id = data[0].id
+
+    db = api.describe_database(database_id=database_id)
+    cols = db.cols
+
+    # resolve columns
+    output_column_id = _resolve_column_name(output_column_name, cols)
+    input_file_column_id = _resolve_column_name(input_file_column_name, cols)
+
```
sg-s, line 42: you don't need to use this anymore -- if you use make_payload from tools.utils, this is done automatically

src/tools/draco.py
```diff
@@ -0,0 +1,70 @@
...
+            },
+        },
+        "outputs": {
+            "output_file": {
+                "rowId": row_id,
+                "columnId": output_column_id,
+                "databaseId": database_id,
+            },
+        },
+        "clusterId": _get_cluster_id(),
```
sg-s, line 59: you also don't need to do this -- this will be filled in automatically

docs/tools/draco.md
vdelannee: To be removed. This is not necessary as we specify input_file_column_name

docs/tools/draco.md
vdelannee: Let ask Max about this

docs/tools/draco.md
vdelannee: OK

docs/tools/draco.md
vdelannee: No, we can rebrand it as it's on MIT License. However, we should include the license in the GitHub repo related to Draco. As the code is not exposed to the customers, I don't think it's really matter, but I am not a lawyer, so I am not sure. At the moment, we are using Decimer and Decimer-segmentation. The last one is going to be changed. The masks are either going to be merged with the one from YOLO, or YOLO is going to replace it.

vdelannee: But, yes we can specify and provide more details regarding the workflow and the tools involved

docs/tools/draco.md
sg-s, line 1: please note that we've now changed how these wrapper functions are exposed -- don't make a new module, but instead put it in the run module

docs/tools/draco.md
bilalshaikh42: This code is actually exposed publicly. We at the minimum need to state that this is a fork and link to the license of the original tool. 
its not just a matter of legal requirements, but opensource etiquette and acklowledgements

vdelannee: Regarding the license, no matter if the code is exposed or not, we should and will include the license. Now, this is not a fork, and we don't share the same visibility, if we do we have a serious problem.
Is only draco.py publicly visible or is there anything outside the client, which is publicly visible?

sg-s:
@aregsyan looks like CI is failing

docs/tools/draco.md
```diff
@@ -0,0 +1,64 @@
...
+
+To start a tool run, use:
+
+```python
+from deeporigin.tools import draco
+
+job_id = draco.start_run(
+    database_id="<your-db-name>",
+    row_id="<row-name>",
+    input_file_column_name="<input-file-column-name>"
```
sg-s, line 37: this needs to change

sg-s (changes requested):

docs/tools/draco.md
```diff
@@ -10,7 +10,7 @@ Upload a PDF file containing chemical structures to extract to the Data Hub.
 
 ## Parameters
 
-This section describes the parameters for a tool run, that are passed in the `start_run` function.
+This section describes the parameters for a tool run, that are passed in the `run_draco` function.
```
sg-s, line 13: there is no run_draco function

sg-s (changes requested):

docs/tools/draco.md
vdelannee: updated

docs/tools/draco.md
vdelannee: This section is not relevant anymore as we don't have any extra parameter. @sg-s start_run is still part of Vina documentation

sg-s: is it?

https://client-docs.deeporigin.io/tools/vina.html#2-start-a-tool-run-on-deep-origin

PR #146, opened 2025-01-28
FEP: tooling to run ABFE

sg-s:
## changes

- [x] ability to run system prep
- [x] pass params transparently 
- [x] docs for abfe
- [x] (unrelated) refactored docs to make them more sensibly organized


src/utils/network.py
```diff
@@ -96,7 +100,7 @@ def _download_nucleus_api_spec() -> None:
 
     spec_file = deeporigin_path / "nucleus_spec.json"
 
-    with open(deeporigin_path / "api_tokens", "r") as file:
+    with open(_get_api_tokens_filepath(), "r") as file:
         data = json.loads(file.read())
```
sg-s, line 104: this is an unrelated change where `api_tokens` was hardcoded

src/tools/utils.py
```diff
@@ -229,7 +229,7 @@ def _column_name_to_column_id(data: dict, cols: list) -> dict:
     Recursively update all values for the key 'columnId' in a nested dictionary.
 
     Args:
-        d (dict): The dictionary to process.
+        data (dict): The dictionary to process.
```
sg-s, line 232: unrelated typo

docs/tools/ref/abfe.md
```diff
@@ -0,0 +1,21 @@
...
+        - force_fields
+        - lig_force_fields
+        - charge_methods
+        - init
+        - complex_prep
+        - emeq
+        - solvation_prep
+        - solvation_fep
+        - simple_md
+        - binding_fep
```
sg-s, line 16: we're ordering them in this way so that it makes sense from the workflow perspective

pyproject.toml
```diff
@@ -58,7 +58,7 @@ dev = [
 docs = [
     "mkdocs", 
     "mkdocs-material-extensions", 
-    "mkdocs-material", 
+    "mkdocs-material>=8.2.0",
```
sg-s, line 61: forces us to use more recent versions of material

PR #150, opened 2025-03-03
Support for new FEP class that can do both ABFE and RBFE

sg-s:


src/tools/fep.py
```diff
@@ -1,14 +1,34 @@
...
+DB_PROTEINS = "Proteins"
+DB_LIGANDS = "Ligands"
+COL_PROTEIN = "Protein"
+COL_LIGAND = "Ligand"
+COL_STEP = "Step"
+COL_JOBID = "JobID"
+COL_OUTPUT = "OutputFile"
+COL_RESULT = "ResultFile"
+COL_DELTA_G = "FEP ΔG (kcal/mol)"
+
```
sg-s, line 29: these constants identify the database that must exist. if it doesn't, it's created on the fly

PR #181, opened 2025-04-07
Implement files.FilesClient interface that talks to our file-service

mike9ant:
This is an early version of files.FileClient interface that talks to our service back-end. It provides methods to download, upload, delete files and sync directory up/down. Client will pick up 'deeporigin config' environment and access token if not overridden.

Current implementation used openapi-python-client to generate the supporting files.file_service module that holds generated wrappers. These are taken from "../../platform/docs/file-service/openapi.yaml". Call scripts/gen_file_service.sh for it.
Right now openapi.yml is patched due to bugs/limitations in server controller.

Most significant files are:
  - src/files/files_client.py  -- core implementation
  - test/files_manual_test_remote.py -- run this script manually to test connection methods by uploading/downloading random files with verification


checkmarx-do:
Logo
**Checkmarx One – Scan Summary & Details** – [88ff0404\-51f0\-4dbf\-bbf2\-0e3028414293](https://us.ast.checkmarx.net/projects/cc182d6a-a56c-4683-afdf-6bd4377eebfb/scans?id=88ff0404-51f0-4dbf-bbf2-0e3028414293&branch=michaels-files-client)
### Great job, no security vulnerabilities found in this Pull Request

pyproject.toml
```diff
@@ -112,7 +112,8 @@ omit = [
 
 [tool.interrogate]
 fail-under = 100
-exclude = ["setup.py", "docs", "build","tests","variables","conftest.py"]
+# Skip files module coverage for now as it has lots of auto-gen code
+exclude = ["setup.py", "docs", "build","tests","variables","conftest.py", "src/files"]
```
sg-s, line 116: we shouldn't exclude this. this check makes sure we have necessary docs

scripts/.openapi-python-client.yml
```diff
@@ -0,0 +1,4 @@
+project_name_override: file-service
+package_name_override: file_service
+package_version_override: "0.1.0"
+use_path_prefixes_for_title_model_names: false
```
sg-s, line 4: this shouldn't be here

scripts/gen_file_openapi.yaml
sg-s, line 1: this shouldn't be here. the client generator can work off a url

scripts/gen_file_service.sh
sg-s, line 1: this shouldn't work like this. 

- it should consume the spec from a URL, not a file
- this should be in a separate repo so that we don't have a bunch of machine-generated code here that users might see

src/files/files_client.py
```diff
@@ -0,0 +1,479 @@
...
+Example usage:
+    ```python
+    from deeporigin.files import FilesClient
+    
+    # Create a client (uses default authentication if token not provided)
+    client = FilesClient()
+
+    # List files in a directory
+    files = client.list_dir('files:///data/', flag="recursive")
+    print(f"Found {len(files)} files")
```
sg-s, line 17: if we use the client pattern (that we use in the platform API and datahub API) we can simplify this even more. instead of methods that require the instantiation of a class, we simply have a function that looks like:

```py

def list_dir(..., client=FilesClient()):
    pass
```

this way, users don't need to know about FilesClient, and this allows us to pass a MockClient object that makes testing easier.

sg-s (changes requested):
high level, i think we should make the following changes:

- all autogenerated code needs to go to a different repo, which will be built into a different package (see platform API or datahub API SDKs)
- here, we merely list that as a dependency. this cuts down on a lot of the machine-generated code, and makes this library cleaner
- having autogen code exist in its own repo allows us to hot-swap SDKs for different versions of the API without large refactors of the code and messy diffs 
- auto gen code should use URLs, not manually downloaded specs

pyproject.toml
mike9ant: the auto-generated code will need to be excluded

scripts/gen_file_openapi.yaml
mike9ant: Not sure about this. Files may be more reliable

src/files/files_client.py
mike9ant: it's inefficient/seems poor design to construct a client every time. in the future we may want to move connection validation checks, API version checks etc into constructor

sg-s: > it's inefficient/seems poor design to construct a client every time.

exactly. that's what i'm suggesting. in the example above, you can call 

```py
list_dir(...)
```

and not pass a client

mike9ant:
@sg-s So I sort-of disagree on separate repo thing. At least it's worth a conversation
Separate repos are hard to track and separate related code from each other

sg-s (approved):

PR #212, opened 2025-05-07
FilesClient fix -- list_folder, creater_folder, delete_folder

mike9ant:
Updated the FilesClient to the latest .yaml interface, fixing list_folder to work with it.
create_folder and delete_folder now work as well.


checkmarx-do:
Logo
**Checkmarx One – Scan Summary & Details** – [30de99a9\-b5b6\-41d9\-b8cd\-05113c6d2174](https://us.ast.checkmarx.net/projects/cc182d6a-a56c-4683-afdf-6bd4377eebfb/scans?id=30de99a9-b5b6-41d9-b8cd-05113c6d2174&branch=michaels-files-update)
### Great job, no security vulnerabilities found in this Pull Request

src/files/files_client.py
```diff
@@ -297,58 +297,65 @@ def _extract_path_from_url(self, url: str) -> str:
...
+        current_iteration = 0
 
-        if flag == "recursive":
-            raise ValueError("list_folder recursive flag not implemented")
+        while current_iteration < max_iterations:
+            call_args = genapi._get_get_object_kwargs(
+                orgFriendlyId=self.organization_id,
+                filePath=remote_path,
+                list_type=2,
+                recursive="true" if recursive == True else None,
```
sg-s, line 324: it's complaining about this because the comparison should be:

```
if recursive
```

src/files/file_service/gen_file_openapi.yaml
```diff
@@ -1,53 +1,107 @@
...
-# This is the OpenAPI specification for the File Service API
-# Had to make fixes to the original generated file to get it to auto-generate correct stubs
-# from this by openapi-python-client.
-# Changes:
-# - Changed the path of get/put/delete to /file-api/{orgFriendlyId}/{filePath}
-# to that it actually passes and argument.
-# - Changed upload to use multipart/form-data instead of application/json
-# - Made list-type as optional, or else server returns JSON instead of file content
+# This is the OpenAPI specification for the File Service API copied here
 openapi: 3.0.0
```
smbatyankh, line 2: Could we include this file in MANIFEST.in? The Deep Origin client installation fails because the file is missing, and adding its path should resolve the issue.

sg-s: i still don't understand why this is here. it should not be

src/files/files_client.py
```diff
@@ -297,58 +297,65 @@ def _extract_path_from_url(self, url: str) -> str:
...
+                continuation_token=continuation_token,
```
sg-s, line 325: first, it looks like continuation_token is always None here.

second, the machinery for all of this seems to simply allow pagination...and this seems like a very unorthodox way of doing it. why can't we simply paginate responses and fetch pages as needed?

src/files/files_client.py
```diff
@@ -377,74 +384,60 @@ def _remote_file_exists(self, path: str) -> bool:
...
-            
+            response = self.client.httpx_request_and_process(call_args)
+
             if response.status_code == 200:
                 return True
             else:
-                self.logger.error(f"Failed to create folder {path}: Status code {response.status_code}")
+                self.logger.error(
+                    f"Failed to create folder {path}: Status code {response.status_code}"
+                )
```
sg-s, line 418: if the user asks to create a folder, and folder creation fails, we should raise an exception

PR #231, opened 2025-05-29
feat: docking now uses UFA

sg-s:
## what

docking now uses UFA to handle files

### DIFF
--- a/src/drug_discovery/docking.py
+++ b/src/drug_discovery/docking.py
@@ -47,7 +47,10 @@ class Docking(WorkflowStep):
         for inputs in job._inputs:
             unique_smiles.update(inputs["smiles_list"])
         num_ligands = len(unique_smiles)
-        return f"Docking <code>{job._metadata[0]['protein_file']}</code> to {num_ligands} ligands."
+
+        protein_file = os.path.basename(job._inputs[0]["protein"]["key"])
+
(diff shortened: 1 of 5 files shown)


src/drug_discovery/docking.py
```diff
@@ -132,160 +142,102 @@ def _get_result_files(
...
-        """return a list of paths to SDF files that contain the poses of all ligands after docking"""
-        return self._get_result_files("ResultFile", ".sdf")
-
-    def _connect(self):
-        """fetch job IDs for this protein from DB"""
-
-        jobs = self._get_jobs()
-        job_ids = [job.id for job in jobs]
-        self._make_jobs_from_ids(job_ids)
-
```
sg-s, line 212: these methods have gone away. we don't need them for reconciliation anymore

src/platform/tools_api.py
```diff
@@ -21,6 +21,8 @@
 TERMINAL_STATES = {"Succeeded", "Failed", "Cancelled"}
 NON_TERMINAL_STATES = {"Created", "Queued", "Running"}
 
+NON_FAILED_STATES = {"Succeeded", "Running", "Queued", "Created"}
```
sg-s, line 24: if jobs are in these states, then we won't re-run things by default

PR #235, opened 2025-06-04
feat: Ability to run RBFE using UFA

sg-s:
## changes

- [x] ability to run rbfe 
- [x] ability to get results
- [x] ability to show results


src/drug_discovery/utils.py
```diff
@@ -78,63 +81,77 @@ def _start_tool_run(
                 + os.path.basename(ligand1_path)
                 + "/"
             )
+        elif tool == "RBFE":
```
aregsyan, line 84: how the paths of ligands and proteins looks like?

PR #295, opened 2025-08-27
fix: client pattern for all functions

sg-s:
# changes

- client pattern, allowing any function to be called with `foo(client=client)`
- docs for client

### DIFF
--- a/src/utils/constants.py
+++ b/src/utils/constants.py
@@ -53,3 +53,5 @@ class PREFIXES:
 
 
 number = int | float
+
+ENVS = Literal["edge", "prod", "staging"]
--- a/docs/dev/clients.md
+++ b/docs/dev/clients.md
@@ -1,0 +1,28 @@
+# Using platform APIs using Deep Origin Platform Client
+
+This document describes how to use the Deep Origin Platform Client. 
+
+## Background
+
+The typical way an end-user would use the Deep Origin python package would be to simply call functions. These functions call various APIs on the Deep Origin platform, using tokens and config information that is read from disk. This approach offers convenience for users who are taking actions as themselves on the platform, within a single organization.  
+
+## Multi-user, multi-org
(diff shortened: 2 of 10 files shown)


src/auth.py
```diff
@@ -25,6 +25,25 @@
...
+}
+
+AUTH_DEVICE_CODE_ENDPOINT = "/oauth/device/code"
+AUTH_TOKEN_ENDPOINT = "/oauth/token"
+AUTH_AUDIENCE = "https://os.deeporigin.io/api"
+
+
+AUTH_GRANT_TYPE = "urn:ietf:params:oauth:grant-type:device_code"
+
+AUTH_CLIENT_ID = {
```
maurocolella, line 41: I see some drawbacks to this, in terms of security (although these are not keys they are half of the story) and maintainability. Is there any way devops could inject these -- and is it worth the effort?

src/auth.py
```diff
@@ -160,15 +183,19 @@ def authenticate() -> dict:
...
     api_refresh_token = response_json["refresh_token"]
 
-    tokens = dict(access=api_access_token, refresh=api_refresh_token)
+    tokens = {
+        "access": api_access_token,
+        "refresh": api_refresh_token,
+    }
 
-    cache_tokens(tokens)
+    if save_tokens:
```
maurocolella, line 191: Same maintainability concern as with the client IDs.

src/utils/constants.py
```diff
@@ -53,3 +53,11 @@ class PREFIXES:
 
 
 number = int | float
+
+ENVS = Literal["edge", "prod", "staging"]
+
+API_ENDPOINT = {
```
maurocolella, line 59: Same issue.

maurocolella (approved):
Some comments about maintainability and future-proofing.

PR #314, opened 2025-09-16
feat: started work on client logging

sg-s:
## changes

the core change here is that we can run the same tests in 3 different modes:

- `record` -- run against live instances, save responses to local sqlite db
- normal --  run against live instances
- `mock` -- run using a mock client, which uses recorded responses

checklist

- [x] mock tests using recorded responses from a mock client
- [x] removed some old and dead code


src/platform/recording.py
```diff
@@ -0,0 +1,183 @@
...
+            override = RequestRecorder(Path(db_path))
+            override.record(
+                method=method, kwargs=kwargs, response=response, duration_ms=duration_ms
+            )
+            return
+
+        request_json = json.dumps(kwargs, sort_keys=True, separators=(",", ":"))
+        request_hash = compute_request_hash(method, kwargs)
+        response_json = json.dumps(response, sort_keys=True, separators=(",", ":"))
+        now_iso = time.strftime("%Y-%m-%dT%H:%M:%S", time.gmtime())
```
maurocolella, line 114: Consider `time.time_ns()` or `datetime.datetime.utcnow().isoformat(timespec="microseconds")` for ordering during rapid tests.

src/platform/recording.py
```diff
@@ -0,0 +1,183 @@
...
+    """SQLite-backed recorder for request/response interactions.
+
+    Thread-safe; maintains a single connection. Designed for tests and lightweight usage.
+    """
+
+    def __init__(self, db_path: Path | None = None):
+        self.db_path = Path(db_path) if db_path else DEFAULT_DB_PATH
+        _ensure_parent_dir(self.db_path)
+        # Use a single connection; enable WAL for better concurrency
+        self._conn = sqlite3.connect(str(self.db_path), check_same_thread=False)
```
maurocolella, line 49: Maybe `.close()` when done.

src/platform/recording.py
```diff
@@ -0,0 +1,183 @@
...
+                cur.execute(
+                    "SELECT MAX(sequence_num) FROM interactions WHERE request_hash=?",
+                    (request_hash,),
+                )
+                max_row = cur.fetchone()
+                max_seq = max_row[0] if max_row else None
+                available = []
+                if max_seq is not None:
+                    available = list(range(0, max_seq + 1))
+                raise KeyError(
```
maurocolella, line 176: Consider a custom exception for clarity.

src/platform/recording.py
```diff
@@ -0,0 +1,183 @@
...
+            method: Fully qualified method name.
+            kwargs: Request arguments.
+            response: Successful response object (JSON-serializable or encodable).
+            duration_ms: Duration of the call in milliseconds.
+            db_path: Optional override DB path for this call.
+        """
+
+        if db_path and Path(db_path) != self.db_path:
+            # If an override is provided, write to that DB lazily via a separate recorder
+            override = RequestRecorder(Path(db_path))
```
maurocolella, line 105: Maybe worth caching in case of heavy use.

src/platform/recording.py
```diff
@@ -0,0 +1,183 @@
...
+            override.record(
+                method=method, kwargs=kwargs, response=response, duration_ms=duration_ms
+            )
+            return
+
+        request_json = json.dumps(kwargs, sort_keys=True, separators=(",", ":"))
+        request_hash = compute_request_hash(method, kwargs)
+        response_json = json.dumps(response, sort_keys=True, separators=(",", ":"))
```
maurocolella, line 113: Consider `try/except` in case the response is malformed.

maurocolella (approved):
Some nits, and in principle I agree with Copilot about md5 vs sha256.

PR #326, opened 2025-09-30
fix: unified job widget using shadow dom

sg-s:
## problem

jupyter lab has an ancient version of bootstrap which messes up any html element using bootstrap. as a result, building html components using bootstrap was a pain/impossible, and things were always broken in jupyter. i had to resort to hacks that worked in jupyter (but didn't work outside of it)

## solution

here, i'm

- namespacing classes
- using shadow dom

so we can use modern bootstrap features out of the box and have it work without any interference from jupyter

## example

Screenshot 2025-09-30 at 11 16 02 AM

## how it looked before

Screenshot 2025-09-30 at 11 18 57 AM


src/templates/job_widget.html
```diff
@@ -0,0 +1,243 @@
...
+                const wrapper = document.createElement('div');
+
+                // Bootstrap CSS inside shadow for isolation
+                const bsLink = document.createElement('link');
+                bsLink.rel = 'stylesheet';
+                bsLink.href = 'https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css';
+                shadow.appendChild(bsLink);
+
+                const style = document.createElement('style');
+                style.textContent = `:host{--bs-border-color:#dee2e6;--bs-card-cap-bg:#f8f9fa}.do-pre,.do-mono{background:#f8f9fa;border:1px solid #e3e3e3;border-radius:.25rem;padding:1rem;font-family:monospace;white-space:pre-wrap}.do-card{margin-bottom:1rem}.do-card-title{font-size:1.5rem;font-weight:500}.do-tabpanel:not(.active){display:none}.card-footer{background-color:var(--bs-card-cap-bg)}.card-footer .badge{padding:.4rem .65rem}`;
```
maurocolella, line 25: Consider breaking this down (even just with line returns) for readability.

src/templates/job_widget.html
```diff
@@ -0,0 +1,243 @@
...
+                                        <td>{{ statuses[i] }}</td>
+                                        <td>{{ started_at[i] }}</td>
+                                        <td>{{ running_time[i] }}</td>
+                                    </tr>
+                                    {% endfor %}
+                                </tbody>
+                            </table>
+                        </div>
+                        <div class="do-tabpanel tab-pane" data-do-panel="inputs">
+                            <pre class="do-pre">{{ inputs_json|safe }}</pre>
```
maurocolella, line 216: Consider using `lang="json"`

maurocolella (approved):
Some very minor nits. Ideally, CSS and JavaScript should be externalized into standalone files for maintainability, but this might not be possible depending on the context. Other than that, LGTM.

src/templates/job_widget.html
sg-s: i don't think this is supported by jinja

PR #328, opened 2025-10-01
feat: support for new terminal states

sg-s:
## changes

- fixed a bug where job widget mixed up IDs
- rolled back to old version of job widget because the new one had weird CSS issues
- support for new terminal states

### DIFF
--- a/src/platform/tools_api.py
+++ b/src/platform/tools_api.py
@@ -16,7 +16,14 @@ __all__ = _add_functions_to_module(
 )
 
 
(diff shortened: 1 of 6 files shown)


src/platform/tools_api.py
```diff
@@ -145,7 +152,7 @@ def cancel_run(
         client=client,
         org_key=org_key,
     )
-    if data["status"] in ["Cancelled", "Failed", "Succeeded"]:
+    if data["status"] in TERMINAL_STATES:
```
aregsyan, line 155: 👍

aregsyan (approved):

PR #470, opened 2026-03-11
feat: (DDOS-4786) add optional parameters for billing, app, session, …

nlinders:
…and project_id across multiple functions


sg-s (commented):
i’m not sure we want to do this. this will require every caller to inject these. instead they should be configured in the client and the client should pass those

PR #522, opened 2026-04-22
Provide projectId in bulk-docking body.

hdabaghyan:
This pull request introduces a small improvement to the `create` method in `src/platform/executions.py` by ensuring that the `projectId` is included in the payload if it is available. This helps maintain consistency and ensures that the execution is correctly associated with the current project.


sg-s (approved):
thanks! good catch

PR #557, opened 2026-05-25
feat: promote DataIngesting status to first-class citizen in SDK

maurocolella:
## Add `DataIngesting` as a first-class execution status

DPS synthesizes a transient `DataIngesting` status while result data is being ingested after tool execution completes. This PR teaches the SDK about it:

- Added to `PlatformStatus` Literal, transition map, and non-terminal/non-failed sets
- Badge renders as light-blue (`info`) with an animated "Ingesting data…" progress bar
- `cancel()` now permitted during `DataIngesting`
- `wait()` and `watch_async()` automatically poll through it (non-terminal)
- DPS mock simulates the status via `metadata.__data_ingestion_started`
- 7 new tests covering polling, cancellation, display, and terminal checks
- Docs updated

### DIFF
--- a/docs/dd/how-to/job.md
+++ b/docs/dd/how-to/job.md
@@ -32,7 +32,7 @@ A dataframe with the following columns will be returned:
 | execution_id | (internal) execution ID. | 
 | completed_at | Time stamp of when the Job was completed |
 | started_at | Time stamp of when the Job was started |
-| status | One of `Succeeded` `Cancelled` `Failed` `Running` `Queued` `Created` `FailedQuotation` `Quoted` or `InsufficientFunds`|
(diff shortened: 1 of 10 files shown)


sg-s (commented):
mauro -- i think there's someting lost in translation -- the client's execution lifecycle (get / sync / wait / watch) polls tools-service (Mason), not data-platform. DataIngesting is synthesized only on the DPS executions projection when deferring terminal upserts during result ingestion- mason never emits it. s oadding DataIngesting to PlatformStatus, terminal sets, and wait/watch behavior is effectively unreachable from current client paths unless we also switch status polling to DPS (or a unified API). imo this looks like dead code relative to how clients actually observe execution state

sg-s:
closing as this has seen no movement

PR #582, opened 2026-06-26
feat: (DDOS-6506) ensure all write paths apply tags across the client

maurocolella:
to do:

- [x] remove `Datasets` 
- [x] remove `tag` and `billing` from every `run` -- pending comments from @andrewmochalskyy


sg-s (approved):

andrewmochalskyy:
### Code review

Found 3 issues (1 bug, 2 tied to tag-management consistency / SDK pass-through):

1. **`changelog` silently dropped** — `Datasets.create()` accepts and documents `changelog`, but the refactored body calls `_build_dataset_admin_set(...)` without it (helper has no such param), so it never reaches the API. On `main` it was sent as `set_dict["changelog"]`. Caller gets no error, silent data loss. (Same applies to `summary`, now only a fallback for `description`.)

https://github.com/deeporiginbio/do-dd-client/blob/ec5aceb76560b0c53fe8532c2b2a8a5c95f0b9e5/src/platform/datasets.py#L210-L233

2. **Dead tag path in molprops (tag-management duplication)** — `Molprops.run()` wraps the body in `with self._execution_tags(tag=tag, billing=billing):`, but the actual create goes through `run_molprops_combined()` → `client.executions.create(...)` directly, never `_create_execution()` (the only reader of those overrides). The context manager is dead code here; tag/billing only work via the explicit kwargs. Drop the wrapper or route through `_create_execution`.

https://github.com/deeporiginbio/do-dd-client/blob/ec5aceb76560b0c53fe8532c2b2a8a5c95f0b9e5/src/drug_discovery/molprops.py#L257-L263

3. **`tags` type changed `list[str]` → `dict` with no coercion (SDK pass-through risk)** — `update_ligand`/`update_protein` changed the `tags` type. There's no beartype on these, so an existing caller (e.g. Balto) passing `tags=["x"]` won't error but will persist a list where the rest of the system now expects a jsonb object. `datasets.create` coerces `list → {"dataset_tags": [...]}`; entities/projects don't. Coerce consistently or keep list back-compat.

https://github.com/deeporiginbio/do-dd-client/blob/ec5aceb76560b0c53fe8532c2b2a8a5c95f0b9e5/src/platform/entities.py#L590-L606

**Nice to have (non-blocking):**

4. `start()` accepts `tag`/`billing` only via `**kwargs` (popped at the body) and documents them, but they aren't in the signature — every other tool method declares them explicitly. Make them real params for discoverability/type-checking.

https://github.com/deeporiginbio/do-dd-client/blob/ec5aceb76560b0c53fe8532c2b2a8a5c95f0b9e5/src/drug_discovery/execution_mixins.py#L126-L131

5. `billing_tag` docstring says `... / DO_TAGS`, but nothing in the client reads a `DO_TAGS` env var — drop the mention or wire it up.

https://github.com/deeporiginbio/do-dd-client/blob/ec5aceb76560b0c53fe8532c2b2a8a5c95f0b9e5/src/platform/client.py#L424-L428

6. `_create_execution`'s `tool_key`/`tool_version`/`timeout` params are never passed by any caller — dead surface, can be removed.

https://github.com/deeporiginbio/do-dd-client/blob/ec5aceb76560b0c53fe8532c2b2a8a5c95f0b9e5/src/drug_discovery/execution.py#L138-L145

(Also: `projects.update()` is new and has no test.)

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/platform/datasets.py
```diff
@@ -153,32 +214,21 @@ def create(
...
+            summary=summary,
+            description=description,
+            source_url=source_url,
+            source_name=source_name,
+            tags=tags,
+            compound_count=compound_count,
+            file_size_bytes=file_size_bytes,
+            dataset_schema=dataset_schema,
+            sample_rows=sample_rows,
+        )
```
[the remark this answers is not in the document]
sg-s: Fixed in a2929bee: `changelog` is forwarded through `_build_dataset_admin_set` as `datasetMeta.changelog` (the bundle shape used by admin create). Mock server flatten updated; `test_create_dataset` asserts persistence.

sg-s:
Addressed in a2929bee: \`Datasets.create(changelog=...)\` now forwards the value via \`datasetMeta.changelog\` in the admin \`set\` bundle (replacing the pre-refactor flat \`set.changelog\` field). Mock flatten + \`test_create_dataset\` updated.

src/drug_discovery/konnektor.py
```diff
@@ -297,6 +297,8 @@ def run(
         *,
         quote: bool = False,
         approve_amount: int | None = None,
+        tag: str | None = None,
+        billing: str | None = None,
```
sg-s: i don't think this is a good pattern -- @andrewmochalskyy @maurocolella 

tag and billing should be owned in the client, not every single tool class

src/drug_discovery/pocket_finder.py
```diff
@@ -268,6 +268,8 @@ def run(
         *,
         quote: bool = False,
         approve_amount: int | None = None,
+        tag: str | None = None,
+        billing: str | None = None,
```
sg-s: see we're repeating this everywhere

src/platform/datasets.py
```diff
@@ -14,6 +14,73 @@
...
+        return {"dataset_tags": tags}
+    return tags
+
+
+def _build_dataset_admin_set(
+    *,
+    name: str,
+    file_path: str,
+    dataset_key: str,
+    dataset_version: str,
```
sg-s: this looks like code to create datasets at the admin level? do we need this in the client?

sg-s (commented):
lots of admin-only code here that i would remove IMO. also @maurocolella plz see pattern of injecting tag and billing (tag?) into every tool class -- this is not sustainable. can we only put it in the client?

tests/test_billing_tag_end_to_end.py
```diff
@@ -8,7 +8,6 @@
 from deeporigin.platform.constants import TOOL_KEYS_AND_VERSIONS
 
 
-@pytest.mark.skip(reason="Temporarily skipped; fix later")
 def test_billing_tag_end_to_end_lv2(client: DeepOriginClient):
     """Run a few molprops tool executions with a unique tag and check billing usage.
```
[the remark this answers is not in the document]
sg-s: this is fine. it's marked lv2

andrewmochalskyy (approved):
Approving — all review items addressed, and `merge_entity_tags` now auto-stamps `app`/`session` provenance on direct entity creates (caller keys win), which is the real fix for the source-filter gap on non-tool paths. Good cleanup (Datasets removed, per-run tag/billing collapsed to client config, coercion unified, `tests/test_tags.py` added).

One heads-up before/at merge — **breaking change**: entity `tags` is now **dict-only** and raises `TypeError` on a list (was `list[str]` on `main` for `update_ligand`/`update_protein`). No callers pass list tags in this repo or the sibling platform repos, so it's safe here — but any out-of-workspace consumer (e.g. a Balto build) passing `tags=[...]` will now throw. Worth a quick ping to the Balto team so they send a dict.

https://github.com/deeporiginbio/do-dd-client/blob/b7f39dfeaea2d93eab14aad76451b90fc7c00b10/src/platform/tags.py#L36-L38

Non-blocking: the SDK now provably writes the `{app, session}` shape, but the true end-to-end (UI source filter returns a directly-added ligand/protein) is still a QA confirmation, not covered by unit tests here.

PR #590, opened 2026-07-16
feat: expose protonation concentration on Ligand

zori-deep-origin:
## Summary

The platform protonation API already returns `concentration_list` alongside `smiles_list` in responses, but `Protonation.run()` was silently discarding it. This PR:

- Adds `protonation_concentration: float | None` field to the `Ligand` dataclass (next to the existing `protonated_at_ph`)
- Reads `concentration_list` from the API response and sets it on each returned `Ligand`
- Gracefully handles responses without `concentration_list` (older API versions)

### Why

We're migrating Balto's ADMET tools from local `molprops` to the platform SDK ([BALTO-245](https://deeporigin.atlassian.net/browse/BALTO-245)). The `ProtonationPredictor` is the last tool to migrate. Balto needs per-species population percentages for its user-facing text output (e.g. "CCO with concentration of 99.93%"). Without this field exposed, Balto would have to dig into raw `responses` to get data the platform already provides.

> **Note:** Balto also renders concentration-vs-pH curves (Henderson-Hasselbalch sweep), which requires pKa values the platform doesn't currently return. That's a separate, larger change — this PR covers the low-hanging fruit that already exists in the response.

## Test plan

- [x] `test_run_exposes_concentration_list` — verifies concentrations are set on both ligands in a multi-state response
- [x] `test_run_handles_missing_concentration_list` — verifies graceful fallback to `None` when key is absent
- [x] `test_run_sets_concentration_on_merged_primary` — verifies the `ligand=` path (merge into primary) also sets concentration

[BALTO-245]: https://deeporigin.atlassian.net/browse/BALTO-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/drug_discovery/protonation.py
+++ b/src/drug_discovery/protonation.py
@@ -191,7 +191,9 @@ class Protonation(Execution, SyncExecutableMixin):
                 inp.mol = new_mol
                 inp.smiles = smi
                 inp.protonated_at_ph = float(self._ph)
-                inp.protonation_concentration = float(conc) if conc is not None else None
+                inp.protonation_concentration = (
+                    float(conc) if conc is not None else None
+                )
                 ligands_out.append(inp)
                 continue
             if n_states == 1:
--- a/src/drug_discovery/protonation.py
+++ b/src/drug_discovery/protonation.py
@@ -176,11 +176,13 @@ class Protonation(Execution, SyncExecutableMixin):
             raise ValueError(
                 "Protonation response missing protonation_states.smiles_list"
             )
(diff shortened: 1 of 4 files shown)


sg-s (approved):
the code is fine, but the tests are written using mock/patch/unittest that we're trying to move away from. if you want to test a platform interaction (in this case, the protonation), use the MockServer (there are other tests with this pattern for reference)

approving if you need to be unblocked quickly

PR #627, opened 2026-09-09
fix(files): always set Content-Length for signed-url uploads (DDOS-7697)

ASinanSaglam:
## Summary

- Always set `Content-Length` on presigned S3 PUTs in `Files._put_to_signed_url`, including `0` for empty files.
- Without it, httpx uses `Transfer-Encoding: chunked`, which S3 presigned URLs reject with `501 NotImplemented`.
- Fixes FEP finalize failures on 0-byte `*_nohup.out` files ([DDOS-7697](https://deeporigin.atlassian.net/browse/DDOS-7697)).
- Unit test asserts empty uploads send `Content-Length: 0`.

**Jira:** [DDOS-7697](https://deeporigin.atlassian.net/browse/DDOS-7697)

## Test plan

- [x] `uv run pytest tests/test_files_signed_url_upload.py --env local`
- [ ] CI green on PR
- [ ] After merge: tag and publish `deeporigin-sdk`; bump pin in md-suite

[DDOS-7697]: https://deeporigin.atlassian.net/browse/DDOS-7697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DDOS-7697]: https://deeporigin.atlassian.net/browse/DDOS-7697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/test_files_signed_url_upload.py
+++ b/tests/test_files_signed_url_upload.py
@@ -74,6 +74,39 @@ def test_put_to_signed_url_refreshes_signed_url_on_retry(
     assert put_attempts["count"] == 2
 
 
+def test_put_to_signed_url_sets_content_length_for_empty_file(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Empty files must send Content-Length: 0 for presigned S3 PUTs."""
+    local_file = tmp_path / "empty.bin"
+    local_file.write_bytes(b"")
+
+    files = _files_with_mock_client()
+
+    class FakeResponse:
+        def raise_for_status(self) -> None:
+            return None
+
+    observed_headers: dict[str, str] = {}
+
+    def fake_put(*_args: object, **kwargs: object) -> FakeResponse:
+        headers = kwargs["headers"]
+        observed_headers["Content-Length"] = headers["Content-Length"]
+        _consume_upload_body(kwargs["content"])
+        return FakeResponse()
+
+    fake_client = MagicMock()
+    fake_client.put.side_effect = fake_put
+    fake_client.__enter__ = MagicMock(return_value=fake_client)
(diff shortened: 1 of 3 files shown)


tests/test_files_signed_url_upload.py
```diff
@@ -74,6 +74,39 @@ def fake_put(*_args: object, **kwargs: object) -> FakeResponse:
...
+        _consume_upload_body(kwargs["content"])
+        return FakeResponse()
+
+    fake_client = MagicMock()
+    fake_client.put.side_effect = fake_put
+    fake_client.__enter__ = MagicMock(return_value=fake_client)
+    fake_client.__exit__ = MagicMock(return_value=False)
+    monkeypatch.setattr(httpx, "Client", MagicMock(return_value=fake_client))
+
+    files._put_to_signed_url(local_file, "/remote/empty.bin", max_retries=0)
```
sg-s: this is fine for now, but i'm trying to move away from magic mocks and monkey patches

sg-s (changes requested):
i think we need a real test (that is run against --env dev) that is marked lv1, where we try to upload a 0-byte file, and it refuses/fails for the right reasons

tests/test_files_signed_url_upload.py
ASinanSaglam: I agree but wasn't sure how well real tests worked in GHA, I'll try to find a better way to tackle it.

sg-s (approved):

PR #630, opened 2026-09-10
feat(drug-discovery): add SecondaryPharmacology SDK class (DDOS-7332)

ASinanSaglam:
## Summary

- Add `SecondaryPharmacology` for `deeporigin.secondary-pharma`: `run()` (served ligand-ml scoring) and `start()` (async docking workflow), gated to their own method since the tool has no `inputs.sync` field to dispatch on
- `get_results()` returns a DataFrame for both paths (`jobOutputs` for ligand-ml, result-explorer for docking, since async completions never populate `jobOutputs`); `get_poses()` (docking-only) builds a downloaded `PoseSet` via `Pose.from_json`
- `self_test` supported on both methods (mutually exclusive with `ligands`); on the docking path neither `get_results()` nor `get_poses()` has anything to return for a self-test run, since the platform publishes no rows for the baked test ligand (no ligand id) — `get_poses()` raises a clear error explaining this
- Mock-server coverage for both paths, docs, and API reference
- Known gaps, not fixed here: the tool-definition enum fetch/validate helpers duplicate `Admet`'s pattern almost verbatim — worth extracting into a shared helper once a third caller needs the same shape; the docking Argo workflow has no ligand batching (`workflow.yaml` is a single task, unlike `deeporigin.docking`'s `batchSize`)

**Jira:** [DDOS-7332](https://deeporigin.atlassian.net/browse/DDOS-7332)

## Test plan

- [x] `uv run pytest tests/test_secondary_pharma.py --env local` (29 passed)
- [x] Live `--env dev` run, both paths: ligand-ml `run()` and full docking `start()` → `wait()` → `get_results()` → `get_poses()` (real Argo workflow, ~110s, 3/3 panel poses downloaded) — confirms the result-explorer read path and `result_type="panelpose"` against the real platform, not just the mock
- [x] Spot-check docs pages for Secondary Pharmacology in the nav

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[DDOS-7332]: https://deeporigin.atlassian.net/browse/DDOS-7332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/dd/tools/secondary-pharma.md
+++ b/docs/dd/tools/secondary-pharma.md
@@ -20,15 +20,15 @@ you which to call instead:
 
 ```mermaid
 flowchart TD
-    ctor["SecondaryPharmacology(method=...)"]
-    ctor -->|" method='ligand-ml' "| ml_run["run()"]
-    ctor -->|" method='docking' "| dock_start["start()"]
+    ctor["SecondaryPharmacology(method=...)"] --> choose{"Choose method"}
+    choose -->|"&nbsp;method='ligand-ml'&nbsp;"| ml_run["run()<br/>served, synchronous scoring"]
+    choose -->|"&nbsp;method='docking'&nbsp;"| dock_start["start()<br/>async Argo workflow"]
 
-    ml_run -->|" completes "| ml_results["get_results()<br/>DataFrame, returned immediately"]
-    ml_run -.->|" start() raises "| blocked(["✗"])
+    ml_run -->|"&nbsp;completes&nbsp;"| ml_results["get_results()<br/>DataFrame, returned immediately"]
+    ml_run -.->|"&nbsp;start() raises&nbsp;"| blocked(("ValueError"))
 
-    dock_start -->|" submits "| dock_wait["wait() / watch()"]
-    dock_start -.->|" run() raises "| blocked
+    dock_start -->|"&nbsp;submits&nbsp;"| dock_wait["wait() / watch()<br/>poll until complete"]
+    dock_start -.->|"&nbsp;run() raises&nbsp;"| blocked
     dock_wait --> dock_results["get_results()<br/>DataFrame, from the platform"]
     dock_results --> dock_poses["get_poses()<br/>downloaded PoseSet"]
 ```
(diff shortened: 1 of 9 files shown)


docs/dd/tools/secondary-pharma.md
```diff
@@ -0,0 +1,126 @@
...
+by `method` at construction:
+
+- `method="ligand-ml"` — served, synchronous XGBoost booster scoring. Use
+  [`run()`](../ref/secondary_pharma.md).
+- `method="docking"` (the default) — an async workflow. Use
+  [`start()`](../ref/secondary_pharma.md).
+
+`run()`, `start()`, and `watch()` are all available on every instance, but
+only the one matching `method` works — the others raise immediately, telling
+you which to call instead:
```
sg-s, line 20: i'd rephrase this to be a little less implementation-specific, and more user friendly....write it as you would like to read it, as a user

src/platform/constants.py
```diff
@@ -166,4 +166,8 @@ def display_platform_status(status: str | None) -> str:
         "tool_key": "deeporigin.import-dataset",
         "tool_version": "latest",
     },
+    "secondary_pharma": {
+        "tool_key": "deeporigin.secondary-pharma",
+        "tool_version": "latest",
```
sg-s: let's pin to major

docs/dd/tools/secondary-pharma.md
ASinanSaglam: Took a second look at the entire docs and reworded more. Tried fixing this initially but didn't go far enough with it, should read better now.

ASinanSaglam:
Pushed the visualization + gap-check surface planned since the last review round:

- `plot()` — heatmap of ligand × target (ligand-ml score, or docking `binding_energy`/`pose_score`)
- `plot_ml_vs_docking()` — `staticmethod`, split heatmap comparing a ligand-ml run against a docking run, sorted so cells both methods agree on collect in the corner
- `get_undocked_ligands()` / `get_missing_pairs()` — docking coverage gap-checks
- `get_results()` now carries a `method` column on every row; ligand-ml results backfill `ligand_id` on read
- Fixed `Execution.list()` (shared base class): no `order=`/`project_id=`, so results came back unordered and spanning the whole org

PR #634, opened 2026-09-16
feat: (DDOS-6754) let the SDK mark a tool run internal

gksdo:
## Summary

The platform accepts a `visibility` tag on tool executions (`"visible" | "hidden"`) that hides a run from user-facing activity views. The SDK had no way to set it. This adds it, client-side only.

This is **not** a user-facing feature — nobody wants to hide their own runs — so it is not exposed on any tool class's `run()` / `start()`. It is for the SDK's own plumbing runs and for internal automation that should not clutter a user's activity.

## Design

- `DeepOriginClient(_visibility="hidden")` stamps every run the client launches, the way `tag` / `billing_tag` do. Underscore-prefixed like `_app` / `_session` to mark it internal, and part of the singleton cache key so two clients differing only in visibility never share an instance.
- `Executions.create(visibility=...)` mirrors the endpoint for per-call use. Precedence: a value in `data`, then the argument, then the client default.
- The winning value is validated against the server enum before any request is made. An explicit `{"visibility": None}` in `data` is dropped rather than sent as `null`.
- `client.py` is staged into `deeporigin-sdk` but `platform/constants.py` is not, so the type is imported under `TYPE_CHECKING` only. Verified by building the wheel and importing it (cf. #631).

## Changes

- `src/platform/constants.py` — `ExecutionVisibility` literal and its value set
- `src/platform/client.py` — `_visibility` on the constructor, the `from_*` factories, and the cache key
- `src/platform/executions.py` — `visibility` on `Executions.create`
- `CONTEXT.md` — glossary entry
- `tests/test_client.py` — 7 tests

## Test plan

- [x] `pytest tests/test_client.py tests/test_executions.py tests/test_tool_execution.py --env local` — 106 pass
- [x] `ruff check` / `ruff format --check`
- [x] `scripts/build_sdk.sh`, install the wheel in a clean venv, import `deeporigin_sdk.platform.client`
- [ ] Launch a run with `_visibility="hidden"` against dev; confirm it is absent from the activity view


andrewmochalskyy (commented):
### Code review

Solid, well-tested change with a clean design (client-level default stamped in `Executions.create`, not on tool classes, matching the DDOS-6506 precedent) and correct handling of the SDK build-staging constraint (`ExecutionVisibility` stays `TYPE_CHECKING`-only in `client.py`). Two things to fix, then two non-blocking notes.

**1. Fix requested: `_visibility` is silently dropped on the no-arg construction path, and three docs promise it isn't.**

The no-arg branch dispatches to `from_env_variables()` / `from_local()` / `from_disk()` with zero arguments and returns before the cache-key line, so `_visibility` (like `_app` / `_session`) is discarded:

https://github.com/deeporiginbio/do-dd-client/blob/f5a2bf2cc6dfab1e30c582dc39dfec71018f4151/src/platform/client.py#L214-L230

Result: `DeepOriginClient(_visibility="hidden")` returns a client whose `_visibility` is `None`, so its runs are **not** hidden. `DeepOriginClient.from_local(_visibility="hidden")` works correctly. The harm is bounded (the flag is lost; there's no cross-contamination, since the shared instance is the plain `None` one), but the new docs assert the opposite guarantee on what the module's own docstrings call the preferred call shape:

https://github.com/deeporiginbio/do-dd-client/blob/f5a2bf2cc6dfab1e30c582dc39dfec71018f4151/src/platform/client.py#L205-L208

https://github.com/deeporiginbio/do-dd-client/blob/f5a2bf2cc6dfab1e30c582dc39dfec71018f4151/src/platform/client.py#L439-L442

https://github.com/deeporiginbio/do-dd-client/blob/f5a2bf2cc6dfab1e30c582dc39dfec71018f4151/CONTEXT.md#L265-L272

The behavior mirrors the pre-existing `_app` / `_session` gap, so how far to take the behavioral fix (thread `_visibility` through the factory dispatch, or raise when an internal-only param arrives with no credentials) is your call. Either way the docs need correcting, since they're wrong as written.

**2. Fix requested: stale cache-key docstring.**

The class docstring still lists the five-field key after `_visibility` was added to the actual tuple (and to the sibling comment and the `Dict[Tuple[...]]` annotation). This one copy was missed:

https://github.com/deeporiginbio/do-dd-client/blob/f5a2bf2cc6dfab1e30c582dc39dfec71018f4151/src/platform/client.py#L295-L299

Actual key:

https://github.com/deeporiginbio/do-dd-client/blob/f5a2bf2cc6dfab1e30c582dc39dfec71018f4151/src/platform/client.py#L241-L244

**3. Non-blocking: client-side enum validation is a hard cross-service coupling.**

Validation is against a Literal that mirrors the server enum, which the comment acknowledges. If the server gains a visibility state before this Literal is updated, the SDK will reject a value the server would accept until a release ships:

https://github.com/deeporiginbio/do-dd-client/blob/f5a2bf2cc6dfab1e30c582dc39dfec71018f4151/src/platform/constants.py#L97-L108

Reasonable given the offline typo-catch benefit; flagging so the coupling is a known, deliberate one.

**4. Non-blocking: `data={"visibility": None}` doesn't suppress a client-level default.**

The precedence generator filters out `None` candidates, so an explicit `None` in `data` is treated as "unset" and falls through to the client-level `_visibility`. A caller passing `data["visibility"]=None` to force server-default on one call still gets the client's hidden default (the workaround is `visibility="visible"`):

https://github.com/deeporiginbio/do-dd-client/blob/f5a2bf2cc6dfab1e30c582dc39dfec71018f4151/src/platform/executions.py#L114-L125

Worth a line in the docstring so the semantics are explicit.

gksdo:
Thanks — both fixes are in `7f72b1e`.

**1. `_visibility` dropped on the no-arg path.** Fixed behaviourally rather than by correcting the docs. You left the call open, and the argument for fixing the behaviour is the flag's own rationale: we validate the value specifically because a lost/mistyped visibility fails *open* — a run meant to be internal quietly shows up in Activity. Silently dropping it on the shape the class docstring calls preferred is that same failure with a different cause, so making the docs true was the better side to fix. `_visibility` is now forwarded into all three factory dispatches, which makes all three doc sites correct as written.

`_app` / `_session` deliberately left alone: that gap predates this flag, and forwarding them would move existing clients between cache keys — not something to slip into this PR.

Tests cover both reachable no-arg branches (local and env-variables) and I confirmed they fail without the fix (`assert None == 'hidden'`), so they actually pin the regression rather than just passing.

**2. Stale cache-key docstring.** Fixed — now lists all six fields.

**4. `data={"visibility": None}` doesn't suppress a client default.** Documented on the `visibility` arg, including the `visibility="visible"` workaround. Left the behaviour as-is: `None` meaning "unset" at every level is what makes the precedence chain uniform, and special-casing an explicit `None` in `data` would make it the one level where `None` means something else.

**3. Enum coupling.** Agreed and deliberate — noting it stays a known tradeoff. If the server gains a visibility state, this Literal needs updating in the same release. Worth a comment on the server enum pointing here; happy to add that if you'd like it in this PR.

`pytest tests/test_client.py tests/test_executions.py tests/test_tool_execution.py --env local` — 109 pass. `ruff check` / `format --check` clean.

nlinders (approved):

sg-s (approved):

andrewmochalskyy (approved):
Both blocking findings from the prior review are resolved with tests (no-arg `_visibility` now forwarded through the factories; cache-key docstring corrected), and a fresh engine pass over the fix commit found no new issues. 109 tests pass. Approving.

